PR #101, opened 2020-04-21
add a 'std' feature that includes serialization and proptest impls

pittma:


Cargo.toml
```diff
@@ -13,6 +13,9 @@ repository = "https://github.com/auxoncorp/ekotrace"
 readme = "README.md"
 build = "build.rs"
 
+[features]
```
ZackPierce, line 16: I think we should explicitly add a `default = []` line here to make it quite clear the that the use of std is bonus rather than baseline.

PR #103, opened 2020-05-04
Compact log reports, chunked and bulk

ZackPierce:
* We have a new trait, `ChunkedReporter` which enables exporting reports in bounded-size chunks.
* The `report` and `report_with_extension` methods have been moved to a distinct `BulkReporter` trait.
* Both chunked and bulk reports encode the compact log (in little-endian form, for consistency) as their primary payload. The header-on-the-wire of the two variants differ slightly due to their distinct use cases.
* The LCM log report format was removed, but the LCM snapshot format remains for now.
* A large pile of incidental fixes and consistency upgrades were made along the way, most notably that LogicalClock.id is now a TracerId rather than a bare u32 and the udp collector uses the ekotrace id types rather than defining its own.


ekotrace-capi/ekotrace-capi-impl/src/lib.rs
```diff
@@ -286,20 +305,151 @@ pub unsafe fn ekotrace_merge_fixed_size_snapshot(
...
+/// The destination buffer pointer must be non-null
+/// and should have an associated size == 256 bytes.
+///
+/// If the function returns any result besides EKOTRACE_RESULT_OK,
+/// there has been an error and the contents of the report destination
+/// are not certain to be initialized or a valid chunk.
+#[cfg_attr(feature = "no_mangle", no_mangle)]
+pub unsafe fn ekotrace_write_next_report_chunk(
+    tracer: *mut Ekotrace<'static>,
+    report_token: *const ChunkedReportToken,
```
mullr, line 403: I'm a little surprised to see the report token being passed around by reference, rather than by value. Any particular reason?

ZackPierce: In rust-land, the token is opaque and created on report-start. Then it can be used by reference for each chunk-write without being duplicated or introspected, and then consumed at report-finish.

ekotrace-capi/include/ekotrace.h
```diff
@@ -42,6 +42,8 @@ typedef struct ekotrace_instant {
     uint32_t event_count;
 } ekotrace_instant;
 
+typedef uint32_t chunked_report_token;
```
mullr, line 45: Is this meant to be the same sized as the Rust structure? It probably is, but that seems really shaky.

ekotrace-capi/ekotrace-capi-impl/src/lib.rs
ZackPierce: The by-default disallowance of token copying in Rust-land is a protective measure to reduce the odds of tokens / reporting sessions getting their wires crossed, so to speak.

ekotrace-capi/include/ekotrace.h
ZackPierce: Yes, that's the point. We can/should add tests for this to be more sure.

ekotrace-capi/include/ekotrace.h
```diff
@@ -298,6 +310,47 @@ size_t ekotrace_merge_fixed_size_snapshot(ekotrace *ekotrace, causal_snapshot *s
...
+ * a value that will be used to produce the
+ * chunks for the report in calls to
+ * `ekotrace_write_next_report_chunk` and `ekotrace_finish_chunked_report`
+ *
+ * Once this method has been called, mutating operations on
+ * the Ekotrace instance will return `EKOTRACE_RESULT_REPORT_LOCK_CONFLICT_ERROR`
+ * until all available chunks have been written with  `ekotrace_write_next_report_chunk`
+ * and `ekotrace_finish_chunked_report` called.
+ */
+size_t ekotrace_start_chunked_report(ekotrace *ekotrace, chunked_report_token * out_report_token);
```
mullr, line 326: `s/* out_report_token/*out_report_token/`

(same below)

ekotrace-capi/include/ekotrace.h
mullr: Yeah, I'd like to see a test for that, as it would be pretty catastrophic were that assumption to fail.

ekotrace-udp-collector/src/chunked.rs
```diff
@@ -0,0 +1,199 @@
...
+            max_total_open_time: chrono::Duration::seconds(600),
+            max_time_since_last_chunk: chrono::Duration::seconds(120),
+        }
+    }
+}
+
+pub struct ChunkHandler {
+    config: ChunkHandlingConfig,
+
+    #[allow(clippy::type_complexity)]
```
mullr, line 22: Man, I'm inclined to agree with clippy here. At least a comment as to the meaning of these different map keys would be welcome.

ekotrace-capi/include/ekotrace.h
ZackPierce: Interestingly, the header size turns out to already be wrong, and the test that exercises it is still passing. Will add static assertions and think about what more to do for testing.

src/compact_log.rs
```diff
@@ -119,40 +160,188 @@ pub(crate) fn split_next_segment(
...
+                            Err(LogEventInterpretationError::EventMissingPayload(event_id))
+                        }
+                        Err(e) => Err(e),
+                    })
+                }
+            } else {
+                Some(
+                    raw_ev
+                        .try_into()
+                        .map_err(|_e| LogEventInterpretationError::InvalidEventId(raw_ev))
```
mullr, line 302: Am I correct in thinking that this is effectively impossible to hit, due to the above bit test?

mullr: (at line 279)

src/report/bulk.rs
```diff
@@ -0,0 +1,202 @@
...
+        extension_metadata: ExtensionBytes<'_>,
+    ) -> Result<usize, ReportError> {
+        let n_log_bytes = self.log.len() * size_of::<CompactLogItem>();
+        let n_extension_bytes = extension_metadata.0.len();
+        if n_log_bytes > core::u32::MAX as usize {
+            // We don't currently support sending reports with logs that big.
+            // N.B. Point for future improvement - actually only send some of the log
+            // by doing some pre-segmentation and determining where we can cut it up,
+            // possibly injecting duplicate clocks as necessary.
+            return Err(ReportError::Encoding);
```
mullr, line 51: This error value doesn't seem right

src/report/bulk.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+        if n_extension_bytes > core::u32::MAX as usize {
+            // We don't support sending reports with extensions that big.
+            // There likely isn't a valid remediation available
+            // since the extension data is opaque.
+            return Err(ReportError::Extension);
```
mullr, line 57: Same here

src/report/bulk.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+        let required_bytes = size_of::<WireBulkHeader>() + n_log_bytes + n_extension_bytes;
+        if destination.len() < required_bytes {
+            return Err(ReportError::InsufficientDestinationSize);
+        }
+        // We consider this relatively safe because WireBulkHeader is a largely
+        // uninterpreted pile of bytes fields all with alignments == 1
+        // and we know from the above check that there's enough size
+        assert_eq_align!(u8, WireBulkHeader);
```
mullr, line 66: This is confusing to me. Isn't /everything/ byte-aligned?

mullr: Of is this asserting that WireBulkHeader doesn't require some coarser alignment?

src/report/bulk.rs
```diff
@@ -0,0 +1,202 @@
...
+    if wire_bytes.len() < size_of::<WireBulkHeader>() {
+        return Err(ParseBulkFromWireError::MissingHeader);
+    }
+    assert_eq_align!(u8, WireBulkHeader);
+    debug_assert_eq!(
+        0,
+        wire_bytes.as_ptr() as usize % align_of::<WireBulkHeader>()
+    );
+    let (header_bytes, payload_bytes) = wire_bytes.split_at(size_of::<WireBulkHeader>());
+    let wire_header = unsafe { &*(header_bytes.as_ptr() as *const WireBulkHeader) };
```
mullr, line 158: While I don't see anything specifically wrong here, this pattern is one that is known to be fragile in many dimensions. Perhaps a binary parser using nom could be more appropriate?

src/report/bulk.rs
ZackPierce: The errors are intentionally sparse and suggestive rather than detailed in this library. 

Is there a particular thing you're hoping to capture? A different extant variant you'd prefer used, or are you asking for a new ReportError variant?

src/report/bulk.rs
ZackPierce: Sanity-check of the latter.

src/report/bulk.rs
ZackPierce: Trying quite hard to keep the dependencies extremely low here. Is there a particular fragility or set thereof you're worried about needing remediation for?

ekotrace-udp-collector/src/chunked.rs
ZackPierce: Righto, will clarify.

ekotrace-capi/include/ekotrace.h
ZackPierce: Will do.

src/compact_log.rs
ZackPierce: Even if the "event with payload" bit is set, doesn't mean that the event is valid, so the prior check on line 279 is not a complete validity pass. For example, the clock bit could also have been set or the underlying raw value was 0.

That said, assuming that the log has been properly created and segmented, I don't expect the highlighted log interpretation errors to surface. The presence of a full validity check via try_into is probably a necessary-but-rare protection against incorrect construction or accidental regressions.

src/report/bulk.rs
mullr: If that's the intent it's fine; it was just a little surprising giving the specificity of the error below.

src/report/bulk.rs
mullr: Dimension 1: Binary protocol / file format parsing code is a very common source of security vulnerabilities in networking code and imaging / video libraries (which are effectively network facing). 

Dimension 2: The practice of coercing a byte buffer into a struct is fragile, as it relies on the struct being laid out in memory just-so. You've done the necessary things to make it so (:picard:), but it relies on compiler behavior that is outside of mainstream. 

Dimension 3: The struct definition itself is a source of fragility; when someone else comes along, it may not be obvious why the struct must be laid out just-so, or that alteingr its contents is verboten. 

The only argument to really be made for struct-coercion is that there is absolutely nothing faster. But in this case, you're already (correctly) going through endian conversion code after the coercion. So I don't think that's what you're going after here, and I don't think that it's warranted in this case anyway. 

Whether we use Nom for it or not, we should probably segregate the wire-protocol code into a separate module with its own (crate-internal) API, where we can easily test the snot out of it with very well-defined boundaries. Definitely round-trip proptests from the API side, and preferably fuzz tests from the binary side.

src/report/bulk.rs
```diff
@@ -0,0 +1,202 @@
...
+    IncompletePayload,
+    /// The data type was not one of the supported varieties
+    UnsupportedDataType(u8),
+    /// The tracer id didn't follow the rules for being
+    /// a valid ekotrace-location-specifying TracerId
+    InvalidTracerId(u32),
+}
+
+// TODO - pick a better fingerprint value
+const BULK_FRAMING_FINGERPRINT_SOURCE: u32 = 2_718_281_828;
```
mullr, line 197: So, this is a magic number indicating 'bulk'? I'd expect to see it as a hex constant in the source, fwiw, since how we usually talk about protocol constants, and it makes it easy to grep for.

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+                    .zip(payload.iter_mut())
+                {
+                    *dest = MaybeUninit::new(CompactLogItem::from_raw(u32::from_le_bytes([
+                        source[0], source[1], source[2], source[3],
+                    ])));
+                }
+                NativeChunk::Log {
+                    header,
+                    contents: NativeChunkLogContents {
+                        n_chunk_payload_items: n_payload_items,
```
mullr, line 437: I think there's a potential consistency problem here, if `n_payload_bytes` is too large on the wire. This is the kind of thing that often causes a buffer overrun as well; I can't find one by visual inspection, but that doesn't convince me it doesn't exist.

ZackPierce: Good point, I can add additional checks in this function, for starters.

src/report/bulk.rs
ZackPierce: Sure, will do.

src/report/bulk.rs
ZackPierce: Definitely agree on even more testing. There are some round-trip proptests over in the allog_log_report bit, but I'm positive there's room for improvement. 

I'll add more, regardless.

As you may have surmised, the main reasons for going with the struct overlay approach initially here were:
* No dependencies required
* Speed of development
* Compiler assistance re: structs as executable specifications. 
* Not much room for unexpected performance costs to hide

 I don't know what to say about "relies on compiler behavior that is outside of mainstream" -- this is a Rust project, and it seems reasonable to rely on the features of the compiler regarding collections of typed fields.  The `repr`s used are stable, and `static_assertion`s help fill in the gaps around comparability, though I'm sure we could add `assert`s to get runtime double-checking around layout if we're feeling suspicious of `static_assertion` as (this project's only real) 3rd party library.

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+    /// Undo any internal mutation locks related to reporting.
+    fn finish_chunked_report(
+        &mut self,
+        token: ChunkedReportToken,
+    ) -> Result<(), ChunkedReportError>;
+}
+
+#[derive(Debug)]
+pub(crate) struct ChunkedReportState {
+    /// Use the None-value to indicate "no chunked report in progress"
```
mullr, line 199: There's no Option here, did you mean 'false'?

ZackPierce: Yep, that's a remnant of a prior version, will correct.

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+    /// Erring on the side of a large
+    /// number of these to support cases like:
+    ///   * a long-lasting partition from the report collector whilst reports
+    /// are archived for future sending
+    ///   * a very high report production cadence relative to the transmission cadence
+    pub chunk_group_id: [u8; 2],
+    /// In what ordinal position does this chunk's payload live relative to the other chunks.
+    pub chunk_index: [u8; 2],
+    /// Does this chunk contain log data (0001) or extension data (0010)?
+    pub chunk_data_type: u8,
```
mullr, line 57: This could also be represented, somewhat more compactly, with a third fingerprint value up-top. Any particular reason to do it in the body here?

ZackPierce: Because the main point of the framing fingerprint is to sanity-check that the following data should be interpreted in a particular way. The interpretation of the header does not change based on the content of the payload.  I think a better name for this field might be "payload_data_type" rather than "chunk_data_type", on further thought, so I understand the conflation.  Will rename.

An additional reason is that we'd like to be able to use the same framing format for log data and extension data, so if the framing is the same, then I think it makes sense to keep the same fingerprint.

Finally, hyper-compaction is not yet a goal, so the nicety of keeping things well aligned won out. Presently the payload at a nice offset (16 bytes in, so also 4 aligned), which is a nice perk for potential optimization later.

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+                let mut payload: [MaybeUninit<CompactLogItem>;
+                    MAX_PAYLOAD_COMPACT_LOG_ITEMS_PER_CHUNK] =
+                    unsafe { MaybeUninit::uninit().assume_init() };
+                let n_payload_items = n_payload_bytes / (size_of::<CompactLogItem>() as u8);
+                for (source, dest) in payload_bytes
+                    .chunks_exact(size_of::<CompactLogItem>())
+                    .zip(payload.iter_mut())
+                {
+                    *dest = MaybeUninit::new(CompactLogItem::from_raw(u32::from_le_bytes([
+                        source[0], source[1], source[2], source[3],
```
mullr, line 431: `source[0..=3]`?

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+    /// The content of the report chunk
+    payload: [MaybeUninit<CompactLogItem>; MAX_PAYLOAD_COMPACT_LOG_ITEMS_PER_CHUNK],
+}
+
+impl PartialEq for NativeChunkLogContents {
+    fn eq(&self, other: &Self) -> bool {
+        self.payload
+            .iter()
+            .take(usize::from(self.n_chunk_payload_items))
+            .zip(
```
mullr, line 518: This will return a false positive if the two have different sizes, but one is a prefix of the other.

src/report/chunked.rs
```diff
@@ -0,0 +1,992 @@
...
+    /// The content of the report chunk
+    payload: [MaybeUninit<u8>; MAX_PAYLOAD_BYTES_PER_CHUNK],
+}
+
+impl PartialEq for NativeChunkExtensionContents {
+    fn eq(&self, other: &Self) -> bool {
+        self.payload
+            .iter()
+            .take(usize::from(self.n_chunk_payload_bytes))
+            .zip(
```
mullr, line 561: Same problem here

src/report/chunked.rs
ZackPierce: `from_le_bytes` accepts a [u8; 4], not a &[u8], so slicing isn't an option here.

src/report/chunked.rs
ZackPierce: Good catch, thank you!

src/report/chunked.rs
ZackPierce: On it!

src/report/bulk.rs
ZackPierce: That gap largely comes from the differences between errors bound for use in the C API versus errors that will likely only ever be exposed in the Rust API.

PR #105, opened 2020-05-20
Impl Arbitray, Serialization traits for TracerId

mullr:


src/id.rs
```diff
@@ -314,7 +314,7 @@ pub mod prop {
 
         fn new_tree(&self, runner: &mut TestRunner) -> NewTree<Self> {
             Ok(EventIdBinarySearch(BinarySearch::new(
-                runner.rng().next_u32(),
```
mullr, line 313: @pittma I think we need at least this to prevent it from putting a zero in there; do we also need to do something in 'simplify'? It seem like the u32 impl we're using would naturally back that off to zero.

src/id.rs
```diff
@@ -323,6 +327,57 @@ pub mod prop {
...
+            self.0.simplify()
+        }
+
+        fn complicate(&mut self) -> bool {
+            self.0.complicate()
+        }
+    }
+
+    #[derive(Debug)]
+    /// A proptest strategy to be used for any valid user event id.
```
ZackPierce, line 359: Typo: valid user event id => valid tracer id

src/id.rs
pittma: > I think we need at least this to prevent it from putting a zero in there

I think that's what I had in mind with the checked add here: https://github.com/auxoncorp/ekotrace/blob/17ce63d2e33f5476b9ab2ff17c34f4e5267e8afc/src/id.rs#L283

As for simplify and complicate... maybe? I think simplify would back it down to 1 thanks to that checked add. As for complicate, I think it'd eventually get stuck at `EventId::MAX_USER_ID`, right?

mullr: oh, I see, or_max has the checked_add in it, and current always goes through that. ok, it's probably good.

PR #111, opened 2020-05-28
debug-collector design doc

NicholasSpringer:


design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+
+In this case, the debug-collector will communicate to the given gdb server over TCP in order to read and write to memory.
+
+### repr(C)
+
+Rust's ABI is undefined, so the representation of structures in memory is unpredictable across compiler versions and architectures. To allow the debug-collector to understand the raw binary it reads from the chip, the ekotrace log data structures must have the `repr(C)` attribute, which tells the compiler to use C's representation of structs in binary. To find the fields of `repr(C)` structs, the debug-collector uses constant offsets for each given word size. This method is fragile to implementation changes to the structs that the debug-collector is accessing.
+
+We would have to add this attribute to `DynamicHistory` and `ChunkedReportState` in ekotrace as well as to `FixedSliceVec` in fixed-slice-vec.
+
+## Symbol resolution
```
jonlamb-gh, line 37: This will likely require the user to put `no_mangle` attributes on their data; are there any Rust tools/libraries out there that we can use to help alleviate this requirement?

design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+
+For each tracer, the user will either pass in the memory location of a pointer to the tracer, or will pass in the name of the symbol of the pointer to the tracer. If the latter is used, then the user would also need to pass in an ELF file containing the symbol table of the program running on the target device. The debug-collector will automatically find the memory location of the given symbol by looping through the given symbol table.
+
+## Locking
```
jonlamb-gh, line 41: This can get tricky in a `no_std` context.
Core does have [AtomicBool](https://doc.rust-lang.org/core/sync/atomic/struct.AtomicBool.html), but as noted in the [Portability Section](https://doc.rust-lang.org/core/sync/atomic/index.html#portability), atomic types are not available on all platforms.
I think often implementations fall back to using `UnsafeCell<T>`, and use `intrinsics::atomic_*`. We might find ourselves doing something similar if we intend on supporting those oddball platforms.
I think for now, saying we'll support platforms that have atomics is probably ok.

design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+
+The debug-collector reads and writes to the ekotrace logs that the tracers running on the device have access to, so the debug-collector must ensure that it is not accessing logs at the same time as tracers. To do so, we can add an `in_use: Atomic<bool>` field to the DynamicHistory struct that will be set to true while the tracer is accessing the log. When the debug-collector pulls the logs, it will halt the CPU of the target and then check the value of `in_use`, aborting if true. If `in_use` is false, then the debug-collector can safely access the logs and then unhalt the CPU when finished. The use of atomic CPU instructions by the `Atomic<bool>` type should prevent race conditions.
+
```
jonlamb-gh, line 44: I think it'd be helpful if you could describe how we intend on performing the "atomic" operation here from the perspective of the debugger.

When the tool/debugger goes to set the `in_use` variable to `true/false`, what instructions/operations will be performed, is it a compare-and-swap/etc?

design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+`ekotrace-debug-collector-cli --session-id 0 --word-size 32 --little-endian --elf ./target-elf --attach stm32 --interval 500 --output-file ./out.csv 0x20001000 TRACER_1 0x20000000 TRACER_3 TRACER_4`
+
+Flags:
+- `--session-id` `-s`: a session id to distinguish the trace data collected during this tracing run
+- `--word-size` `-w`: processor word size of the target system. If not given, detected from elf file if possible. Defaults to 32 with warning.
+- `--little-endian` `-le`; `--big-endian` `-be`: endianness of target system. If not given, detected from elf file if possible. Defaults to little with warning.
+- `--elf` `-e`: ELF file to use for symbol resolution and/or architecture detection
+- `--attach` `-a`: chip type passed to probe-rs
+- `--gdb-server` `-g`: address of gdb server attached to chip
+- `--interval` `-i`: interval between log report attempts
```
jonlamb-gh, line 61: Can we make flexible, something like [parse_duration](https://docs.rs/humantime/2.0.0/humantime/struct.Duration.html).

design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+- `--output-file` `-o`: a path to the CSV file where the data will be logged
+
+
+# Open Questions
+
+- Is probe-rs compatible with a wide enough variety of devices? (unclear to me whether it works with word sizes other than 32-bit)
```
jonlamb-gh, line 67: I think this is ok for now.

design/debug-collector.md
```diff
@@ -0,0 +1,68 @@
...
+- Other than elf files, what situations regarding target binaries will we need to handle?
```
jonlamb-gh, line 68: ELF files seem fine to me. We could consider others, like Intel HEX for example, but really we're after the symbols and we're not likely to get them with most of the other formats.

design/debug-collector.md
NicholasSpringer: By tool/debugger, do you mean the debug-collector itself? I was originally proposing that it would halt the CPU in order to prevent tracers from entering the mutually exclusive area, so it wouldn't write to the in_use variable at all. It would just check in_use to make sure that when the CPU is halted, the tracer is not already in the middle of a log access.

jonlamb-gh: Ah I see, I misread and didn't catch the halt-and-read part.

design/debug-collector.md
```diff
@@ -0,0 +1,127 @@
...
+
+Invariants:
+- In the output vector, every entry is at the index that the writer wrote it at. For instance, the 20th entry that the writer wrote would be the 20th element of the output vector. However, since data in the buffer may be overwritten before the reader reads it, any number of the entries in the output vector may be replaced by a nil value which represents that the entries were missed. 
+- For two-word entries (like clock entries or entries with a payload), the algorithm guarantees that each two word entry will either be fully included or fully omitted in the ouput vector; there will never be half of a two-word entry included.
+
+#### Writer Algorithm
+
+The writer keeps track of a write cursor, which represents the index of the next write. This cursor is incremented every time a write happens, and is not reset when the end of the backing storage is reached. Therefore, the actual index of the backing storage is calculated as: `write_cursor % len_of_storage`.
```
jonlamb-gh, line 63: Since we're targeting embedded systems, I think it would be reasonable to constrain the ring buffer size to a power of 2. That way we can get rid of the expensive modulo operator and instead use bitwise arithmetic.

NicholasSpringer: Yeah, that seems reasonable

design/debug-collector.md
```diff
@@ -0,0 +1,127 @@
...
+
+When reading, the reader also must check if both of the last two values of its temporary read buffer are nil instead of just the last. If a reader reads a prefix of a two-word entry but then misses the suffix, then the prefix will be replaced by a nil value in the output vector.
+
+#### Next steps
+
+A prototype is fully implemented, and has no known issues. It holds up against random timouts between a large number of reads and writes. More thorough testing of the prototype could be done by adding random timeouts to the internals of the data structure in order to target more extreme edge cases. Beyond that, we will probably need to do model checking to verify that the buffer works in all cases.
+
+#### Changes to ekotrace
+
+This data structure would replace the current structure for holding compact log entries. The RaceBufferReader requires the std library, but adding functionality to read from the writer structure itself would be trivial if concurrency is not necessary or if locking systems are available. Adding this functionality would be necessary so that log reports can be created if a UDP collector is in use.
```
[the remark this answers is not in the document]
ZackPierce: 👍

PR #112, opened 2020-05-28
Configurable CLI manifest-gen as a library

jonlamb-gh:
Includes the commit from #108 to get CI checks.


jonlamb-gh:
I had gone that route originally but found that most of the code would have to move since there are shared types across the various subcommands, so I stuck with providing all of the cli things via a `[lib]` inside the cli crate. Our other product(s) using the cli library currently depend on most of the available types.

PR #113, opened 2020-06-02
Add C API const qualifiers

jonlamb-gh:


ekotrace-capi/src/lib.rs
```diff
@@ -229,15 +229,11 @@ pub fn ekotrace_default_panic_abort(_info: &core::panic::PanicInfo) -> ! {
     // to pull in libc_print for operating systems that support it.
     // A separate alternative would be to provide a hook for
     // setting a panic-handler implementation at runtime.
-    unsafe {
-        core::intrinsics::abort();
-    }
+    core::intrinsics::abort();
```
ZackPierce, line 232: What happened to make this okay to call without an `unsafe` block?  [The documentation](https://doc.rust-lang.org/core/intrinsics/fn.abort.html) still lists this function as unsafe.  (Same question for below)

src/history.rs
```diff
@@ -539,20 +538,12 @@ impl<'a> DynamicHistory<'a> {
             if external_clock.count == 0 {
                 continue;
             }
-            let id: TracerId = match external_clock.id.try_into() {
```
ZackPierce, line 542: Good catch that this check isn't strictly necessary anymore.

ekotrace-capi/src/lib.rs
jonlamb-gh: Ran across this while cleaning up the warnings:

```text
warning: unnecessary `unsafe` block
   --> src/lib.rs:232:5
    |
232 |     unsafe {
    |     ^^^^^^ unnecessary `unsafe` block
    |
    = note: `#[warn(unused_unsafe)]` on by default

warning: unnecessary `unsafe` block
   --> src/lib.rs:240:5
    |
240 |     unsafe {
    |     ^^^^^^ unnecessary `unsafe` block

warning: 2 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
```

I haven't chased this one down yet, wonder if we should instead decorate them with a `allow(unused_unsafe)` for now?

ZackPierce: I believe I found the relevant change. https://github.com/rust-lang/rust/pull/72204. Hasn't hit stable, but c'est la vie.  No change necessary here.

PR #122, opened 2020-06-10
Add collected-log-to-graph builder library, use to export graph in CLI

### LINKED ISSUE #114: Rethinking the ekotrace analysis CLI
## Overview
Rather than having `analysis` as a subcommand, Ekotrace should have an `export` subcommand:

```
ekotrace export <input_file.ekt>  [--type=cyclic | --type=acyclic default:acyclic] --locations=something.csv --events=other.csv [-o,--output <output_file> default: out.<format>] [--segments-only] [--format=dot,gexf,gml default: dot]
```

This subcommand produces 4 different representations of the causal history given by `input_file`, and those are selected by the arguments given via the `--type` switch, and whether or not `--segments-only` is used.

&nbsp; | Cyclic | Acyclic 
-|-|-
**With `--segments-only`** | A topology graph, with edges between tracers weighted by occurrence. | The causal history on a segment-by-segment basis. |
**Without  `--segments-only`** | A state graph, with edges between event types weighted by occurrence. | The complete causal history, event-by-event. |

These graphs will include the metadata for the events and the tracers in a form that is amenable to tools like Gephi. For `dot` output types, for instance, this information will go in the `attrs` section of nodes and edges.

While the CLI documented above states that there will be more than one format available, for the initial implementation, we'll stick to just dot.

## Changes

1. The `analysis` subcommand is replaced with this new export command.

## Implementation details

In order to legibly "address" the vertices in our dot files, we need to use metadata we have available, but are prevented from doing this directly in Petgraph, as it identifies vertices with their `NodeIndex`s. Instead, the graph building code shall use the same API (`add_node` / `add_edge`) to construct simple vertex and edge lists, this way, we can be parametric over which graph library is being used by the code that ingests the streamed, unordered columnar (CSV at the time of writing) data and converts it into a graph, leaving it possible to have the data in a Petgraph graph should we wish to make use of its analysis.

The other area we may wish to be parametric is the incoming data source. This would allow for the raw-data-set-to-graph-form code to be reusable with any backend storage, so long as it may be treated as a stream (or very long) list of `Event`s.

pittma:
This PR also adds the `export` subcommand that uses `ekotrace-graph` to
build different graph-based views of a session's collected reports.

---

Closes #114.


nathanaschbacher:
Per our conversion in Slack about changes to the new export data:

1. We need to put a graph identifier before the `{` in `digraph {`, like `digraph G {`
2. For the DG-style graphs we need to make edge `occurrence` be `weight` for the number of times we observe that edge being traversed.
3. For the DG-style graphs we need to add a `weight` attribute to the Nodes that is the number of times we've seen that event.
4. For the DAG-style `--interactions-only` graph we need to add a `weight` attribute that indicates how many events were in that segment, even though we don't actually display/keep every discrete event. Because this can still be useful metadata to provide context for what's going on with communication in the system (i.e. especially if there are "lumpy" segments where the sizes vary radically). 
5. We need to make sure that the attribute types are `""`ed when they're string-y and not-`""` when they're numbers.
6. On the DAG-style graphs rename the `clock` attribute to `tracer_epoch`
7. On the DAG-style graphs that include all the events rename the `clock_offset` attribute to `epoch_offset`
8. Everywhere we use "location" use "tracer" instead (e.g. "raw_location_id" -> "raw_tracer_id").

nathanaschbacher:
9. On the `--interactions-only` exports (both DG and DAG) accumulate the tags for all the events within the segment (akin to accumulating the number of events in the segment) into a superset of all the tags for all the events observed and put it into a `tags` attribute.

ekotrace-graph/src/meta.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/// A row in tracers.csv for a component.
+#[derive(Debug, Clone, Deserialize)]
+pub struct TracerMeta {
+    pub id: u32,
+    pub name: String,
+    pub description: String,
+    pub file: String,
+    pub function: String,
```
jonlamb-gh, line 41: We recently removed the function field from the manifests.

PR #130, opened 2020-06-12
RaceBuffer implementation

NicholasSpringer:
Includes implementation of RaceBuffer as ekotrace sub-crate, but does not implement use of RaceBuffer into ekotrace internals.

### DIFF
--- a/race-buffer/src/lib.rs
+++ b/race-buffer/src/lib.rs
@@ -152,12 +152,11 @@ where
 
         // Calculate number of entries in snapshot buffer that may have been overwritten
         let post_wcurs = self.snapper.snap_wcurs();
-        let overlap =
-            if post_wcurs > self.storage_cap + first_read {
-                (post_wcurs - self.storage_cap) - first_read
-            } else {
-                0
-            };
+        let overlap = if post_wcurs > self.storage_cap + first_read {
+            (post_wcurs - self.storage_cap) - first_read
+        } else {
+            0
+        };
         if overlap >= buf_snapshot.len() {
             // All entries may have been overwritten, return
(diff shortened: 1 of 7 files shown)


race-buffer/src/lib.rs
```diff
@@ -0,0 +1,421 @@
+#![cfg_attr(not(feature = "default"), no_std)]
```
ZackPierce, line 1: The usual idiom in Rust is to have a feature named "std" that, when present, means that the standard library should be in use.

race-buffer/src/lib.rs
```diff
@@ -0,0 +1,421 @@
...
+    cursor & (storage_cap - 1)
+}
+#[repr(C)]
+/// Struct used to write to buffer
+pub struct RaceBuffer<'a, T, H>
+where
+    T: core::marker::Copy + core::cmp::PartialEq,
+    H: Fn(T) -> bool,
+{
+    wcurs: usize, // Writer's cursor
```
mullr, line 16: These should be formatted like this:
```rust
struct Foo {
  /// comment
  x: u32
}
```
To work well with rustdoc

race-buffer/src/lib.rs
```diff
@@ -0,0 +1,420 @@
...
+    S: Snapper<E>,
+{
+    /// Struct for reading writer state
+    snapper: S,
+    /// Reader's cursor
+    rcurs: usize,
+    /// Capacity of backing storage
+    storage_cap: usize,
+    phantom: PhantomData<E>,
+}
```
NicholasSpringer, line 115: Is this a reasonable use of PhantomData to get this struct to compile?

PR #133, opened 2020-06-16
Debug collector input processing

NicholasSpringer:
Processing of CLI input into a config that can immediately be used by the collector to collect data.
#124, #127


NicholasSpringer:
> I don't think we want to be tracking binaries in this repo, can those two bins be generated when needed by the tests instead?

Yes, but then those tests would require linkers for specific architectures.

jonlamb-gh:
> > I don't think we want to be tracking binaries in this repo, can those two bins be generated when needed by the tests instead?
> 
> Yes, but then those tests would require linkers for specific architectures.

We should be able to use rust-lld I would think.

NicholasSpringer:
> We should be able to use rust-lld I would think.

Oh, ok. I'll look into that

ekotrace-debug-collector/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ekotrace-debug-collector"
+version = "0.1.0"
+authors = ["Nicholas Springer <nicholas.m.springer@gmail.com>"]
```
jonlamb-gh, line 4: I think we've all been using our Auxon addresses in here, `nspringer@auxon.io`?

ekotrace-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,549 @@
...
+            }
+        );
+        assert_eq!(
+            config_from_options(options_from_str(
+                "ekotrace-debug-collector \
+                --session-id 0 \
+                --attach stm32 \
+                --interval 1s \
+                --output-file ./out \
+                --elf ./tests/example-be-elf \
```
NicholasSpringer, line 483: I switched to a runtime-compiled crate for testing the symbol parsing, but I think it makes sense for endianness detection to just include a small binary. It's just 600 bytes and basically just consists of the ELF header that is read to figure out endianness, so there's not any reason we'd want to change this.

jonlamb-gh:
I think we can close this in favor of #142

PR #141, opened 2020-06-23
Make error mod more accessible to library consumers

jonlamb-gh:
* `pub` the error module for library users
* Use the error pattern consistently in the subcommands
* Remove the logging prefix configuration now that the product names are aligned

### DIFF
--- a/modality-probe-cli/src/main.rs
+++ b/modality-probe-cli/src/main.rs
@@ -1,4 +1,4 @@
-use modality_probe_cli::{export, header_gen, manifest_gen, opts::*};
+use modality_probe_cli::{error::GracefulExit, export, header_gen, manifest_gen, opts::Opts};
 use structopt::StructOpt;
 
 fn main() {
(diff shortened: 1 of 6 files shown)


modality-probe-cli/src/main.rs
```diff
@@ -12,10 +12,6 @@ fn main() {
     match opts {
         Opts::ManifestGen(opt) => manifest_gen::run(opt),
         Opts::HeaderGen(opt) => header_gen::run(opt, internal_events),
-        Opts::Export(exp) => {
-            if let Err(e) = export::run(exp) {
-                println!("Error: {}", e);
-            }
-        }
+        Opts::Export(exp) => export::run(exp).unwrap_or_exit("export"),
```
pittma, line 15: This loses the error message I think. I think that's a thing we'd probably want to hold on to.

jonlamb-gh: I'm not seeing anything being dropped, does this look sane to you?
The message is coming from [export.rs](https://github.com/auxoncorp/ekotrace/blob/05b3dfead23bd9634f6a6ea2f8691809370af58a/modality-probe-cli/src/export.rs#L59):

```text
cargo run -- export cyclic --events ev --probes pr --report re
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `.../target/debug/modality-probe export cyclic --events ev --probes pr --report re`
modality-probe export: error: failed to open the probes file at pr: No such file or directory (os error 2)
```

pittma: Ah, crap you're right. I'd just cursorily looked and apparently only paid attention to the `GracefulExit` impl for `Option<T>`.

Issue #125, opened 2020-06-10
Construct debug-collector internal code from prototype


NicholasSpringer:
- Structure internal code so it can be used for the cli
- Design/implement timed collection
- Implement use of RaceBufferReader

PR #142, opened 2020-06-23
Debug collector internal code

NicholasSpringer:
Issue #125 
Internal code for collecting logs over JTAG/SWD. Contains git history of #133.


modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,618 @@
...
+)]
+pub struct CLIOptions {
+    /// Session id to associate with the collected trace data
+    #[structopt(short = "s", long = "session-id", default_value = "0")]
+    session_id: u32,
+
+    /// Specifies little endian architecture of target system
+    #[structopt(short = "l", long = "little-endian")]
+    little_endian: bool,
+    /// Specifies big endian architecture of target system
```
jonlamb-gh, line 55: minor nit: spacing looks different here

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,618 @@
...
+    name = "modality-probe-debug-collector",
+    about = "Periodically collects logs from microcontrollers over debug interfaces; outputs them to a file."
+)]
+pub struct CLIOptions {
+    /// Session id to associate with the collected trace data
+    #[structopt(short = "s", long = "session-id", default_value = "0")]
+    session_id: u32,
+
+    /// Specifies little endian architecture of target system
+    #[structopt(short = "l", long = "little-endian")]
```
jonlamb-gh, line 53: I think you can add another config option here to lean more on StructOpt: `conflicts_with = "big-endian"`

modality-probe-debug-collector/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+modality-probe = { path = ".." }
+util = { path = "../util" }
+modality-probe-udp-collector = { path = "../modality-probe-udp-collector" }
+
+[dependencies.race-buffer]
+path = "../race-buffer"
+features = ["std"]
+
+
```
jonlamb-gh, line 23: nit: extra spacing here

modality-probe-debug-collector/.gitignore
```diff
@@ -0,0 +1 @@
+tmp
```
jonlamb-gh, line 1: This isn't needed if we use [tempfile](https://crates.io/crates/tempfile).

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,672 @@
...
+
+    /// Chip type of target system to attach to
+    #[structopt(short = "a", long = "attach")]
+    attach_target: Option<String>,
+
+    /// Address of gdb server to connect to
+    #[structopt(short = "g", long = "gdb-server")]
+    gdb_addr: Option<SocketAddrV4>,
+
+    /// Interval between log collections
```
jonlamb-gh, line 71: I think this opt doc line should include the expected format

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,672 @@
...
+    use super::*;
+    use std::format;
+    use std::fs::canonicalize;
+    use std::process::Command;
+    use std::str::FromStr;
+    use std::time::Duration;
+
+    const LE_SYMBOLS_BIN_PATH: &str =
+        "./tests/symbols-example/target/thumbv7em-none-eabihf/debug/symbols-example";
+    const BE_BIN_PATH: &str = "./tmp/be-example-bin";
```
jonlamb-gh, line 237: I think we should be using [tempfile](https://crates.io/crates/tempfile) here.

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,672 @@
...
+
+    fn options_from_str(input: &str) -> CLIOptions {
+        CLIOptions::from_iter(input.split(" "))
+    }
+
+    /// Test basic parsing
+    #[test]
+    fn test_basic() {
+        assert_eq!(
+            config_from_options(options_from_str(
```
jonlamb-gh, line 305: StructOpt provides an iterator pattern for this:
```rust
Opts::from_iter(["modality-probe-debug-collector", "--session-id", "..."].iter())
```

modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1142 @@
...
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+
+// NOTE: These may be changed once RaceBuffer is implemented into ekt
+const PROBE_ID_OFFSET: u32 = 0x0;
+const BUF_ADDR_OFFSET: u32 = 0x4;
+const BUF_CAP_OFFSET: u32 = 0x8;
+const WCURS_OFFSET: u32 = 0xc;
+//const _EVENT_COUNT_OFFSET: u32 = 0x0;
+//const _CLOCKS_ADDR_OFFSET: u32 = 0x0;
+//const _CLOCKS_LEN_OFFSET: u32 = 0x0;
```
jonlamb-gh, line 32: Are these commented out lines still relevant?

src/compact_log.rs
```diff
@@ -64,18 +64,21 @@ impl CompactLogItem {
         )
     }
 
+    /// prototype
```
jonlamb-gh, line 67: What does "prototype" mean here?

modality-probe-debug-collector/src/cli.rs
NicholasSpringer: That's what I'm using to generate the options from the string. Is there a problem with formatting the input as a string literal?

modality-probe-debug-collector/src/lib.rs
NicholasSpringer: I wasn't sure if they would be needed later, but I'll delete them for now.

modality-probe-debug-collector/src/cli.rs
jonlamb-gh: Ah I see. Nope, I just missed that helper method, seems fine.

modality-probe-debug-collector/src/main.rs
```diff
@@ -0,0 +1,10 @@
+use structopt::StructOpt;
+
+mod cli;
+use cli::{config_from_options, CLIOptions};
+
+fn main() {
+    let opts = CLIOptions::from_args();
+    let config = config_from_options(opts);
+    println!("Running debug collector with configuration: {:#?}", config);
```
jonlamb-gh, line 9: I don't think I would expect to see this type of output unless some sort of `--verbose` flags was passed.

modality-probe-debug-collector/tests/symbols-example/src/main.rs
```diff
@@ -0,0 +1,29 @@
+//! Blinks an LED
```
jonlamb-gh, line 1: Is this true? I don't see any LED blinky code.

src/id.rs
```diff
@@ -198,6 +198,8 @@ impl EventId {
     /// neighbors that attempt to communicate with it.
     pub const EVENT_NUM_CLOCKS_OVERFLOWED: EventId =
         EventId(unsafe { NonZeroU32::new_unchecked(EventId::MAX_INTERNAL_ID - 4) });
+    /// Raw nil value used by RaceBuffer. Must not be used as an event id elsewhere
+    pub const EVENT_NIL_VALUE_RAW: u32 = EventId::MAX_INTERNAL_ID - 5;
```
jonlamb-gh, line 202: This event doesn't match the type of the internal events, should be `EventId`.

src/id.rs
```diff
@@ -198,6 +198,8 @@ impl EventId {
...
 
     /// The events reserved for internal use
     pub const INTERNAL_EVENTS: &'static [EventId] = &[
```
jonlamb-gh, line 205: Should your extra event be in this list?
So it shows up as an internal event in the generated event manifests, or is it intentionally excluded?
I can't remember if a user would ever see this event in their log?

NicholasSpringer: I originally had it raw to use it as a constant for a CompactLogItem trait. I just changed it to an EventId and made some functions const to be able to use it that way. The user will never see the EVENT_NIL_VALUE in the output log, but I included it in the INTERNAL_EVENTS anyway.

NicholasSpringer: Actually, as I understand, a list of internal events with descriptions is exposed to the user, so I don't think it makes sense to include it in `INTERNAL_EVENTS`

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,666 @@
...
+    /// Symbols and/or addresses of probes or probe pointers
+    probe_syms: Vec<String>,
+}
+
+/// Turn CLI options into configuration for the collector
+pub(crate) fn config_from_options(options: CLIOptions) -> Result<Config, Box<dyn Error>> {
+    let mut probe_addrs = Vec::new();
+    let mut symbols = Vec::new();
+    // Separate probe/probe pointer addresses and probe symbols
+    for addr_str in options.probe_syms.iter().filter(|s| s.starts_with("0x")) {
```
ZackPierce, line 94: Naive question -- do we want/need to support addresses prefixed with `0X` as well as `0x` ?

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,666 @@
...
+        probe_addrs,
+    })
+}
+
+/// Open elf file for parsing
+fn open_elf<'a>(path: &PathBuf, elf_buf: &'a mut Vec<u8>) -> Result<Box<Elf<'a>>, Box<dyn Error>> {
+    let mut file = File::open(path)?;
+    file.read_to_end(elf_buf)?;
+    match Elf::parse(elf_buf) {
+        Ok(elf) => Ok(Box::new(elf)),
```
ZackPierce, line 154: IIUC, we don't need to be boxing this Elf object, and can instead just return it directly (with appropriate update to function signature).

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,666 @@
...
+        Err(err) => Err(Box::new(err)),
+    }
+}
+
+/// Get value and size of given symbol, or None if symbol cannot be found
+fn parse_symbol_info(elf_file: &Elf, symbol_name: &str) -> Option<(u32, u32)> {
+    let log_sym = elf_file.syms.iter().find(|sym| -> bool {
+        let name_opt = elf_file.strtab.get(sym.st_name);
+        if let Some(name_res) = name_opt {
```
ZackPierce, line 163: This double-nested destructure could be done in a single pass for clarity:

```rust
if let Some(Ok(name)) = name_opt {
    name == symbol_name
} else {
    false
}
```

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,666 @@
...
+                }
+            }
+        } else {
+            // Default to little endian with warning
+            println!("Warning: Endianness not specified, using little endian");
+            Ok(false)
+        }
+    } else if o.little_endian {
+        if let Some(elf_endianness) = elf_endianness_opt {
+            if let Ok(endianness) = elf_endianness {
```
ZackPierce, line 200: Similar collapsing here can prevent rightward drift:

```rust
if let Ok(Endian::Big) = elf_endianness {
   println!("...warning...");
}
Ok(false)
```

modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,666 @@
...
+                output_path: "./out".into(),
+                probe_addrs: vec![ProbeAddr::Addr(0x100)]
+            }
+        )
+    }
+
+    /// Should error if both attach target and gdb server specified
+    #[test]
+    #[should_panic]
+    fn error_on_multiple_targets() {
```
ZackPierce, line 349: I generally discourage the use of `should_panic` because it makes it more likely for the wrong sort of error tossed early to masquerade as a desired panic later in a function. 

Clearly that's less pressing of a concern for one-liners like this, but as a general practice, I would recommend doing a pattern of:

```rust
assert!(operation_returning_a_result().is_err());
```
which works simply enough when you don't have precise error management,
but can be upgraded in a straightforward manner when specific error variants
are exposed, e.g.

```rust
assert_eq!(Err(MyFancyErr:DesiredVariant)), operation_returning_a_result());
```

modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1139 @@
...
+use modality_probe::{EventId, LogicalClock, ProbeId};
+use modality_probe_udp_collector::add_log_report_to_entries;
+use race_buffer::reader::{RaceBufferReader, Snapper};
+use race_buffer::Entry;
+use util::alloc_log_report::LogReport;
+use util::model::{LogEntry, SegmentId, SessionId};
+
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+
+// NOTE: These may be changed once RaceBuffer is implemented into ekt
```
ZackPierce, line 25: Probably ought to document how these values are determined -- e.g. where to go look in source to update them.

modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1139 @@
...
+impl Snapper<CompactLogItem> for MemorySnapper {
+    fn snap_wcurs(&self) -> Result<usize> {
+        Ok(self.mem_reader.borrow_mut().read32(self.wcurs_addr)? as usize)
+    }
+
+    fn snap_storage(&self, index: usize) -> Result<CompactLogItem> {
+        let raw: u32 = self
+            .mem_reader
+            .borrow_mut()
+            .read32(self.storage_addr + 4 * (index as u32))?;
```
ZackPierce, line 85: Where does this `4` constant come from?  Naming or commenting would be a help here.

modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1139 @@
...
+struct MemorySnapper {
+    /// Reader used to read device memory
+    mem_reader: Rc<RefCell<dyn MemoryReader>>,
+    /// Address of RaceBuffer backing storage
+    storage_addr: u32,
+    /// Address of RaceBuffer write cursor
+    wcurs_addr: u32,
+}
+
+impl Snapper<CompactLogItem> for MemorySnapper {
```
ZackPierce, line 76: Because this trait has zero doc-comments or discussion in the design doc, it is tricky to determine its intent and whether the implementation at hand is doing the right thing.

src/id.rs
```diff
@@ -198,6 +198,9 @@ impl EventId {
     /// neighbors that attempt to communicate with it.
     pub const EVENT_NUM_CLOCKS_OVERFLOWED: EventId =
         EventId(unsafe { NonZeroU32::new_unchecked(EventId::MAX_INTERNAL_ID - 4) });
+    /// Raw nil value used by RaceBuffer. Must not be used as an event id elsewhere
+    pub const EVENT_NIL_VALUE: EventId =
+        EventId(unsafe { NonZeroU32::new_unchecked(EventId::MAX_INTERNAL_ID - 5) });
```
ZackPierce, line 203: If we simply must have a NIL value event, I would like to request it to be at the edge of our internal range (and just bump down the others) rather than having it be at an unexpected offset.

src/history.rs
```diff
@@ -72,6 +72,7 @@ const_assert_eq!(
 /// backed by runtime-sized arrays of current logical clocks
 /// and probe log items
 #[derive(Debug)]
+#[repr(C)]
```
ZackPierce, line 75: This is a step in the right direction for sorting out a fixed layout. `FixedSliceVec` will need to be updated to be `repr(C)` for this to be complete due to its use in the `clocks` and `compact_log` fields. All of the other fields are `repr(C)` or have transparent primitive layouts.

src/id.rs
ZackPierce: In case it wasn't clear, I would rather *not* have such an event, if there's a clean way around it.

src/compact_log.rs
```diff
@@ -102,6 +105,14 @@ impl core::fmt::Debug for CompactLogItem {
     }
 }
 
+impl race_buffer::Entry for CompactLogItem {
+    const NIL_VAL: Self = CompactLogItem::event(EventId::EVENT_NIL_VALUE);
+
+    fn is_prefix(&self) -> bool {
```
ZackPierce, line 111: I'm worried about naive, log-context-free interpretation of this method.  The reason is that both the "count" and the "payload" log items may very well have those bits set.  

Without correct alignment of log interpretation (e.g. awareness of log item position relative to a stable start-point), it's possible to mistake one of the payload log items for one of the "prefixy" items.

modality-probe-debug-collector/src/lib.rs
NicholasSpringer: I will add doc-comments to the trait

src/history.rs
NicholasSpringer: Yes, `FixedSliceVec` will need to become `repr(C)`, but I was holding off on dealing with that until I update DynamicHistory, which is not in the scope of this PR.

src/compact_log.rs
NicholasSpringer: I was actually not aware that the count and payload items were allowed to use those bits, but the RaceBuffer guarantees that the reader never reads the value after a prefix if the prefix is missed. I will do more model checking to ensure that this doesn't cause problems.

NicholasSpringer:
I removed a good amount of code from this PR that will eventually be replaced with generic log processing, and left commented code in the places that the generic log processing code will be called.

PR #145, opened 2020-06-29
Reduce snapshot overhead

### LINKED ISSUE #116: Reducing snapshot overhead to encourage putting it in all component communcations.
@jonlamb-gh has noticed while instrumenting various systems with Ekotrace that the most effective way to build up the causal history between two different tracers is if the snapshot data is co-located with the wire and/or IPC communication. This of course makes sense, as OoB snapshot exchange based on the information that's currently being traced would only capture the causality of the snapshot exchange itself.

On the one hand, we could modify the data that we're tracing to attempt to make it so OoB snapshot exchange has enough data to allow us to reconstruct causal history once all the log reports have been received, but doing this would require the send-side to locally record that it attempted to send some message relative to its local clock, and the receive-side would have to record that it received the message relative to its local clock, and we'd have to be able to uniquely identify every message, and record that in the tracer's data so that we could connect the dots for which message was sent when and received when in the respective sender's and receiver's event histories. 

This seems fairly intractable to make low overhead as both the sender and receiver would need a way to deterministically identify the message the same way (like content hashing it or something) to be able to reconstruct its notion of the message's unique identifier before recording it. This also would balloon the size of the local trace data to store this additional information. (NOTE: I could see this mode of tracking as being something to opt-in to where you get to decide if you want to enable tracing "transparently" at the cost of storage and computation in exchange for not having to pack our opaque bytes onto the user's existing data). There are other caveats that are significant here too, like collisions in how one computes uniqueness (e.g. sending the same message multiple times) or things that can happen normally on the wire that alter computed uniqueness relative to semantic uniqueness (e.g network hardware rewriting metadata as it moves through a network).

On the other hand, we can try to reduce the overhead of Ekotrace snapshots and provide library support to make it trivial for users to pack our opaque bytes into their communication streams and unpack & merge them upon receipt on the other end. This strikes me as the much more attractive option, but I'm open to alternative viewpoints.

With respect to reducing the overhead, one suggestion offered by @jonlamb-gh and @mullr is removing the constant overhead of the extension data indicator/metadata, so that we can get the entirety of a snapshot to fit in 8 bytes.

### LINKED ISSUE #143: Remove LCM

jonlamb-gh:
This commit changes the representation of a causal snapshot to
a fixed 12 byte history with only the local probe's logical clock.

Closes #116
Closes #143


src/history.rs
```diff
@@ -262,8 +251,10 @@ impl<'a> DynamicHistory<'a> {
         self.increment_local_clock_count();
         self.write_current_clocks_to_log();
         Ok(CausalSnapshot {
-            probe_id: self.probe_id.get_raw(),
-            clock: self.clocks[0].count,
+            clock: LogicalClock {
```
ZackPierce, line 254: This could probably become `clock: self.clocks[0],`

src/lib.rs
```diff
@@ -44,22 +43,33 @@ pub struct CausalSnapshot {
...
     /// Construct a causal snapshot from a sequence of little endian bytes
-    pub fn from_le_bytes(words: [u32; 3]) -> Self {
-        let res_lsb = words[2] & core::u16::MAX as u32;
-        let res_msb = (words[2] >> 16) & core::u16::MAX as u32;
-        CausalSnapshot {
-            probe_id: words[0],
-            clock: words[1],
-            reserved_0: res_lsb as u16,
-            reserved_1: res_msb as u16,
+    pub fn from_le_bytes(words: [u32; 3]) -> Result<Self, ModalityProbeError> {
```
ZackPierce, line 46: There is only one error outcome possible in this function, so this error type seems way wide. The `InvalidProbeId`  type seems like a better fit.

jonlamb-gh: :+1: 91b6bb9122bdee6bd6578fd965b91f36190cda1f

src/history.rs
jonlamb-gh: :+1:

PR #149, opened 2020-06-29
Replace the wire types

jonlamb-gh:
This commit replaces the fixed-size always-initialized wire structs
with read/write wrapper types over bytes.

### DIFF
--- a/src/wire/bulk_report.rs
+++ b/src/wire/bulk_report.rs
@@ -1,7 +1,6 @@
 //! A wire protocol for representing Modality probe log reports in bulk form
 
-use crate::ProbeId;
-use byteorder::{ByteOrder, LittleEndian};
+use crate::{wire::le_bytes, ProbeId};
 
 /// Everything that can go wrong when attempting to interpret a bulk report
 /// from the wire representation
@@ -137,14 +136,14 @@ impl<T: AsRef<[u8]>> BulkReport<T> {
     #[inline]
     pub fn fingerprint(&self) -> u32 {
         let data = self.buffer.as_ref();
-        LittleEndian::read_u32(&data[field::FINGERPRINT])
+        le_bytes::read_u32(&data[field::FINGERPRINT])
     }
 
     /// Return the `probe_id` field
     #[inline]
     pub fn probe_id(&self) -> Result<ProbeId, BulkReportWireError> {
         let data = self.buffer.as_ref();
-        let raw_probe_id = LittleEndian::read_u32(&data[field::PROBE_ID]);
+        let raw_probe_id = le_bytes::read_u32(&data[field::PROBE_ID]);
         match ProbeId::new(raw_probe_id) {
             Some(id) => Ok(id),
             None => Err(BulkReportWireError::InvalidProbeId(raw_probe_id)),
@@ -155,14 +154,14 @@ impl<T: AsRef<[u8]>> BulkReport<T> {
     #[inline]
     pub fn n_log_bytes(&self) -> u32 {
         let data = self.buffer.as_ref();
-        LittleEndian::read_u32(&data[field::N_LOG_BYTES])
+        le_bytes::read_u32(&data[field::N_LOG_BYTES])
     }
 
     /// Return the `n_extension_bytes` field
     #[inline]
     pub fn n_extension_bytes(&self) -> u32 {
         let data = self.buffer.as_ref();
-        LittleEndian::read_u32(&data[field::N_EXT_BYTES])
+        le_bytes::read_u32(&data[field::N_EXT_BYTES])
     }
 }
 
@@ -181,28 +180,28 @@ impl<T: AsRef<[u8]> + AsMut<[u8]>> BulkReport<T> {
     #[inline]
     pub fn set_fingerprint(&mut self) {
         let data = self.buffer.as_mut();
-        LittleEndian::write_u32(&mut data[field::FINGERPRINT], Self::FINGERPRINT);
+        le_bytes::write_u32(&mut data[field::FINGERPRINT], Self::FINGERPRINT);
     }
 
     /// Set the `probe_id` field
     #[inline]
     pub fn set_probe_id(&mut self, value: ProbeId) {
         let data = self.buffer.as_mut();
-        LittleEndian::write_u32(&mut data[field::PROBE_ID], value.get_raw());
+        le_bytes::write_u32(&mut data[field::PROBE_ID], value.get_raw());
     }
 
     /// Set the `n_log_bytes` field
     #[inline]
     pub fn set_n_log_bytes(&mut self, value: u32) {
         let data = self.buffer.as_mut();
-        LittleEndian::write_u32(&mut data[field::N_LOG_BYTES], value);
+        le_bytes::write_u32(&mut data[field::N_LOG_BYTES], value);
     }
 
     /// Set the `n_extension_bytes` field
     #[inline]
     pub fn set_n_extension_bytes(&mut self, value: u32) {
         let data = self.buffer.as_mut();
-        LittleEndian::write_u32(&mut data[field::N_EXT_BYTES], value);
+        le_bytes::write_u32(&mut data[field::N_EXT_BYTES], value);
     }
 
     /// Return a mutable pointer to the payload
--- a/src/wire/bulk_report.rs
+++ b/src/wire/bulk_report.rs
@@ -1,0 +1,349 @@
+//! A wire protocol for representing Modality probe log reports in bulk form
+
+use byteorder::{ByteOrder, LittleEndian};
+
+/// Everything that can go wrong when attempting to interpret a bulk report
+/// from the wire representation
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum BulkReportWireError {
+    /// The fingerprint didn't match expectations
+    InvalidFingerprint,
+    /// There weren't enough bytes for a full header
+    MissingHeader,
+    /// There weren't enough payload bytes (based on
+    /// expectations from inspecting the header).
+    IncompletePayload,
+}
+
+/// A read/write wrapper around a bulk report buffer
+#[derive(Debug, Clone)]
+pub struct BulkReport<T: AsRef<[u8]>> {
+    buffer: T,
+}
+
+mod field {
+    type Field = ::core::ops::Range<usize>;
+    type Rest = ::core::ops::RangeFrom<usize>;
+
+    /// A magical (constant) value used as a hint about the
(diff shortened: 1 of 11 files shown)


src/wire/bulk_report.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn payload(&self) -> &'a [u8] {
+        let data = self.buffer.as_ref();
+        &data[field::PAYLOAD]
+    }
+}
+
+impl<T: AsRef<[u8]> + AsMut<[u8]>> BulkReport<T> {
+    /// Set the `fingerprint` field
+    #[inline]
+    pub fn set_fingerprint(&mut self, value: u32) {
```
ZackPierce, line 173: I don't think we have any reason to ever use a value other than FINGERPRINT, so I recommend we remove the `value` argument.

src/wire/bulk_report.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    /// Construct a bulk report from a byte buffer, with checks.
+    ///
+    /// A combination of:
+    /// * [new_unchecked](struct.BulkReport.html#method.new_unchecked)
+    /// * [check_len](struct.BulkReport.html#method.check_len)
+    /// * [check_fingerprint](struct.BulkReport.html#method.check_fingerprint)
+    /// * [check_payload_len](struct.BulkReport.html#method.check_payload_len)
+    pub fn new_checked(buffer: T) -> Result<BulkReport<T>, BulkReportWireError> {
```
ZackPierce, line 60: I think the idiom is for the checked version to just be called `new`.

src/wire/bulk_report.rs
```diff
@@ -0,0 +1,349 @@
...
+}
+
+impl<T: AsRef<[u8]>> AsRef<[u8]> for BulkReport<T> {
+    fn as_ref(&self) -> &[u8] {
+        self.buffer.as_ref()
+    }
+}
+
+impl<T: AsRef<[u8]> + AsMut<[u8]>> AsMut<[u8]> for BulkReport<T> {
+    fn as_mut(&mut self) -> &mut [u8] {
```
ZackPierce, line 214: I don't recall - why do we want to expose the raw bytes mutably? For alternate mutable overlays? Some other reason?

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    /// Construct a chunked report from a byte buffer, with checks.
+    ///
+    /// A combination of:
+    /// * [new_unchecked](struct.ChunkedReport.html#method.new_unchecked)
+    /// * [check_len](struct.ChunkedReport.html#method.check_len)
+    /// * [check_fingerprint](struct.ChunkedReport.html#method.check_fingerprint)
+    /// * [check_payload_len](struct.ChunkedReport.html#method.check_payload_len)
+    pub fn new_checked(buffer: T) -> Result<ChunkedReport<T>, ChunkedReportWireError> {
```
ZackPierce, line 92: Similar naming nit as above -- could just be `new`

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    pub fn payload(&self) -> &'a [u8] {
+        let data = self.buffer.as_ref();
+        &data[field::PAYLOAD]
+    }
+}
+
+impl<T: AsRef<[u8]> + AsMut<[u8]>> ChunkedReport<T> {
+    /// Set the `fingerprint` field
+    #[inline]
+    pub fn set_fingerprint(&mut self, value: u32) {
```
ZackPierce, line 233: Similar recommendation re: not permitting possible-invalid fingerprints to be set.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+}
+
+impl<T: AsRef<[u8]>> AsRef<[u8]> for ChunkedReport<T> {
+    fn as_ref(&self) -> &[u8] {
+        self.buffer.as_ref()
+    }
+}
+
+impl<T: AsRef<[u8]> + AsMut<[u8]>> AsMut<[u8]> for ChunkedReport<T> {
+    fn as_mut(&mut self) -> &mut [u8] {
```
ZackPierce, line 302: Similar query as before -- are we sure we need `as_mut`? I think it would be safer to require external users that want to muck with the underlying bytes to deconstruct/unwrap this instance then re-wrap it (optionally with checking) after they're done.

jonlamb-gh: Yeah I think your right.
I thought it be a convenience thing, but after thinking about it again, seems more like a :foot: :gun:.
I'll get those removed.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Return the `chunk_index` field
+    #[inline]
+    pub fn chunk_index(&self) -> u16 {
+        let data = self.buffer.as_ref();
+        LittleEndian::read_u16(&data[field::CHUNK_INDEX])
+    }
+
+    /// Return the `payload_data_type` field
+    #[inline]
+    pub fn payload_data_type(&self) -> u8 {
```
ZackPierce, line 194: Seems like this ought to return `ChunkPayloadDataType` if this wrapper's job is to do interpretation.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Set the `chunk_index` field
+    #[inline]
+    pub fn set_chunk_index(&mut self, value: u16) {
+        let data = self.buffer.as_mut();
+        LittleEndian::write_u16(&mut data[field::CHUNK_INDEX], value);
+    }
+
+    /// Set the `payload_data_type` field
+    #[inline]
+    pub fn set_payload_data_type(&mut self, value: u8) {
```
ZackPierce, line 261: I would expect the value to be `ChunkPayloadDataType` and let this wrapper handle the wire-conversion.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+        let data = self.buffer.as_mut();
+        data[field::PAYLOAD_DATA_TYPE] = value;
+    }
+
+    /// Set the `is_last_chunk` field
+    #[inline]
+    pub fn set_is_last_chunk(&mut self, value: u8) {
```
ZackPierce, line 268: Seems like the value ought to be `bool` here.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Return the `payload_data_type` field
+    #[inline]
+    pub fn payload_data_type(&self) -> u8 {
+        let data = self.buffer.as_ref();
+        data[field::PAYLOAD_DATA_TYPE]
+    }
+
+    /// Return the `is_last_chunk` field
+    #[inline]
+    pub fn is_last_chunk(&self) -> u8 {
```
ZackPierce, line 201: Seems like we should be returning `bool` here to encourage semantic consistency.

src/wire/bulk_report.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Set the `fingerprint` field
+    #[inline]
+    pub fn set_fingerprint(&mut self, value: u32) {
+        let data = self.buffer.as_mut();
+        LittleEndian::write_u32(&mut data[field::FINGERPRINT], value);
+    }
+
+    /// Set the `probe_id` field
+    #[inline]
+    pub fn set_probe_id(&mut self, value: u32) {
```
ZackPierce, line 180: Recommend accepting `value: ProbeId` instead.

src/wire/chunked_report.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// Set the `fingerprint` field
+    #[inline]
+    pub fn set_fingerprint(&mut self, value: u32) {
+        let data = self.buffer.as_mut();
+        LittleEndian::write_u32(&mut data[field::FINGERPRINT], value);
+    }
+
+    /// Set the `probe_id` field
+    #[inline]
+    pub fn set_probe_id(&mut self, value: u32) {
```
ZackPierce, line 240: Recommend accepting `value: ProbeId` instead.

.github/workflows/cross_compile.yml
```diff
@@ -17,11 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: nightly
```
ZackPierce, line 20: As mentioned on slack, we should probably stay on stable for CI to avoid accidental drift into nightly-required.

src/wire/causal_snapshot.rs
```diff
@@ -0,0 +1,200 @@
...
+
+/// Everything that can go wrong when attempting to interpret a causal snaphot
+/// from the wire representation
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub enum CausalSnapshotWireError {
+    /// There weren't enough bytes for a full causal snapshot
+    MissingBytes,
+    /// The probe id didn't follow the rules for being
+    /// a valid Modality probe-specifying ProbeId
+    InvalidProbeId(u32),
```
ZackPierce, line 15: It seems this case is only relevant for when we're reading bytes, yet this type is used for both read and write.

src/wire/causal_snapshot.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/// A read/write wrapper around a causal snaphot buffer
+#[derive(Debug, Clone)]
+pub struct CausalSnapshot<T: AsRef<[u8]>> {
```
ZackPierce, line 20: Seems odd to have a wire wrapper type with the same name as the canonical in-memory representation type.

src/wire/causal_snapshot.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    /// Consumes the causal snaphot, returning the underlying buffer
+    pub fn into_inner(self) -> T {
+        self.buffer
+    }
+
+    /// Return the length of a buffer required to hold a causal snaphot
+    pub fn buffer_len() -> usize {
```
ZackPierce, line 78: This method name seems confusing, because it doesn't actually return the buffer's len, it returns an expectation about what a buffer's len should at least be. A careful reader will note the lack of a `self` parameter, but perhaps for clarity we could make it a `const` function with a slightly more explicit name. Maybe `min_buffer_len` ?

jonlamb-gh: The name change makes sense to me. 
I held off on making the associated methods `const` to avoid `feature(const_fn)`, I think it's still unstable:
> error[E0723]: trait bounds other than `Sized` on const fn parameters are unstable

PR #160, opened 2020-07-06
Move some things around to reduce clutter at the root of the repo

### LINKED ISSUE #151: Top-level crate refactoring
In preparation for open sourcing there's a few things that need to be done for aesthetic purposes.

- [x] Move `util::model` and `alloc_log_report` into root `modality-probe` crate behind the existing `std` feature gate.
- [x] Factor out log assembly and csv writing so it can be reused across collectors.
- [x] Move graph builder crate into CLI (at least until it's actually needed for something else).

pittma:
Closes #151.


debian-package/Cargo.toml
```diff
@@ -13,9 +13,7 @@ name = "debian_package"
 path = "src/lib.rs"
 
 [dependencies]
-modality-probe-cli = { path = "../modality-probe-cli" }
 modality-probe-capi = { path = "../modality-probe-capi" }
-modality-probe-udp-collector = { path = "../modality-probe-udp-collector" }
```
jonlamb-gh, line 18: Is there any way to keep these deps?
This is how `cargo-deb` automatically produces the Debian package `Depends` (runtime dependencies).
I think this change would require us to manage the `Depends` manually.

PR #162, opened 2020-07-07
remove use of reserved nil values from race buffer

NicholasSpringer:
Update race buffer crate and TLA+ model to remove the need for reserved nil values. This was necessary because clock counts and payload items can take any u32 value, so it is impossible to reserve any CompactLogItem value as nil.

### DIFF
--- a/race-buffer/model/Makefile
+++ b/race-buffer/model/Makefile
@@ -24,7 +24,7 @@ tla2tools.jar:
 	(grep -q "\-\-algorithm" $*.tla && ${PLUSCAL} $*.tla) || true
 
 	@ echo "\n\n-------------- TLC ------------------"
-	${TLC} -workers ${WORKERS} -metadir metadir/check/$* $*.tla 2>&1 | tee $@.tmp
+	${TLC} -deadlock -workers ${WORKERS} -metadir metadir/check/$* $*.tla 2>&1 | tee $@.tmp
 	@ echo "-------------------------------------\n\n"
 	@! grep "Error:" $@.tmp
 	mv $@.tmp $@
--- a/race-buffer/model/RaceBuffer.cfg
+++ b/race-buffer/model/RaceBuffer.cfg
@@ -4,7 +4,7 @@
 
 CONSTANTS
     BufCapacity = 4
-    NumWrites = 8
+    NumWrites = 10
 
 INIT Init
 NEXT Next
@@ -12,5 +12,10 @@ NEXT Next
 \* PROPERTY
 \* Uncomment the previous line and add property names
 
-INVARIANT InvCorrectIndeces InvConsistentDoubles
+INVARIANT 
+    InvRcursBehindWcurs
+    InvWcursBehindOwcurs
+    InvReadBufferNotTooBig
+    InvCorrectIndeces 
+    InvConsistentDoubles
 \* Uncomment the previous line and add invariant names
--- a/race-buffer/model/RaceBuffer.tla
+++ b/race-buffer/model/RaceBuffer.tla
@@ -6,7 +6,7 @@ CONSTANTS
 
   NumWrites  (* Number of writes to execute *)
 
-ASSUME Assumption == BufCapacity \in Nat\*(Nat \ 0..2) (* Buffer capacity is at least 3 *)
+ASSUME Assumption == BufCapacity \in (Nat \ 0..2) (* Buffer capacity is at least 3 *)
 
 -----------------------------------------------------------------------------
 
@@ -294,7 +294,7 @@ InvStoredPrefixNotSuffix ==
   storedPrefix.type = "NONE" \/ storedPrefix.type = "PREFIX"
 
 (* Check that every non-nil entry is at the correct index in the read buffer *)
-InvCorrectIndeces == 
+InvCorrectIndices == 
   IF Size(rbuf) > 0 
   THEN \A i \in 0..(Size(rbuf)-1): rbuf[i].index = i \/ rbuf[i].type = "MISSED"
   ELSE TRUE
--- a/race-buffer/model/RaceBuffer.tla
+++ b/race-buffer/model/RaceBuffer.tla
@@ -6,111 +6,93 @@ CONSTANTS
 
   NumWrites  (* Number of writes to execute *)
 
-ASSUME Assumption == BufCapacity \in (Nat \ 0..2) (* Buffer capacity is at least 3 *)
+ASSUME Assumption == BufCapacity \in Nat\*(Nat \ 0..2) (* Buffer capacity is at least 3 *)
 
 -----------------------------------------------------------------------------
 
 VARIABLES 
   wpc, (* Writer's program counter *)
   buf, (* Backing storage *)
-  wcurs, (* Write cursor - Equal to number of entries already written *)
-
-  writes, (* Number of entries written (used to limit state space) *)
+  wcurs, (* Write cursor - Points to next write position *)
+  owcurs, (* Overwrite cursor - Marks where space is allocated for the writer to write freely *)
 
   rpc, (* Reader's program counter *)
   rbuf, (* Buffer of entries already read *)
   bufSnap, (* Intermediate buffer used to store a snapshot of entries being read *)
-  rcurs, (* Read cursor - Equal to number of entries already read*)
-  wcursSnap (* Value of wcurs observed at beginning of read *)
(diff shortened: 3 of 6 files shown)


race-buffer/src/lib.rs
```diff
@@ -2,227 +2,37 @@
...
-use core::marker::PhantomData;
-
-#[cfg(not(feature = "std"))]
-use core::fmt;
-#[cfg(feature = "std")]
-use std::fmt;
 
 pub trait Entry: Copy + PartialEq {
-    const NIL_VAL: Self;
+    //const NIL_VAL: Self;
```
jonlamb-gh, line 7: Probably don't need this commented line anymore?

race-buffer/src/lib.rs
```diff
@@ -2,227 +2,37 @@
...
-        self.storage[get_cursor_index(self.storage.len(), curs)]
-    }
-
-    /// Get current value of write cursor
-    pub fn read_wcurs(&self) -> usize {
-        self.wcurs
+fn get_cursor_index(storage_cap: usize, cursor: usize, use_base_2_indexing: bool) -> usize {
+    if use_base_2_indexing {
+        // Index is lowest n bits of cursor where storage_cap is 2^n
+        cursor & (storage_cap - 1)
```
jonlamb-gh, line 16: If your not already testing for this elsewhere, might be nice to have an extra debug check here:
`debug_assert!(is_power_of_two(storage_cap));`

race-buffer/src/lib.rs
```diff
@@ -232,18 +42,19 @@ pub mod tests {
     pub struct OrderedEntry(u32);
 
     impl Entry for OrderedEntry {
-        const NIL_VAL: OrderedEntry = OrderedEntry(0);
+        //const NIL_VAL: OrderedEntry = OrderedEntry(0);
```
jonlamb-gh, line 45: Probably can axe this one too?

race-buffer/src/lib.rs
```diff
@@ -268,62 +79,97 @@ pub mod tests {
...
             self.0 >= Self::SUFFIX_TAG && self.0 < Self::PREFIX_TAG
         }
 
         // Invariant: Check if entries all have correct index
-        fn entries_correct_index(rbuf: &[OrderedEntry]) -> bool {
+        fn entries_correct_index(rbuf: &[Option<OrderedEntry>]) -> bool {
             for i in 0..rbuf.len() {
-                if rbuf[i].to_raw() != i as u32 + 1 && rbuf[i] != Self::NIL_VAL {
-                    return false;
+                if let Some(entry) = rbuf[i] {
```
jonlamb-gh, line 93: I think `filter_map` can be used here to collapse things a bit:

```rust
fn entries_correct_index(rbuf: &[Option<OrderedEntry>]) -> bool {
    for (idx, entry) in rbuf.iter().filter_map(Option::some).enumerate() {
        if entry.to_index() != idx as u32 + 1 {
            return false;
        }   
    }       
    return true;
}
```

race-buffer/src/lib.rs
```diff
@@ -268,62 +79,97 @@ pub mod tests {
...
-                let next = &rbuf[i + 1];
-                if (cur.is_prefix() && !next.is_suffix()) || (!cur.is_prefix() && next.is_suffix())
-                {
+        fn double_entries_consistent(rbuf: &[Option<OrderedEntry>]) -> bool {
+            if let Some(first_entry) = rbuf[0] {
+                if first_entry.is_suffix() {
                     return false;
                 }
             }
+            if let Some(last_entry) = rbuf[rbuf.len() - 1] {
```
jonlamb-gh, line 109: The Rust idiom for this is [last()](https://doc.rust-lang.org/std/primitive.slice.html#method.last).

race-buffer/src/lib.rs
NicholasSpringer: I'll add an explicit test for this

race-buffer/model/RaceBuffer.cfg
```diff
@@ -4,13 +4,18 @@
...
 
 \* PROPERTY
 \* Uncomment the previous line and add property names
 
-INVARIANT InvCorrectIndeces InvConsistentDoubles
+INVARIANT 
+    InvRcursBehindWcurs
+    InvWcursBehindOwcurs
+    InvReadBufferNotTooBig
+    InvCorrectIndeces
```
mullr, line 19: 'Indices'

race-buffer/model/RaceBuffer.tla
```diff
@@ -6,111 +6,93 @@ CONSTANTS
 
   NumWrites  (* Number of writes to execute *)
 
-ASSUME Assumption == BufCapacity \in (Nat \ 0..2) (* Buffer capacity is at least 3 *)
+ASSUME Assumption == BufCapacity \in Nat\*(Nat \ 0..2) (* Buffer capacity is at least 3 *)
```
mullr, line 9: Stale comment here?

race-buffer/model/Makefile
```diff
@@ -24,7 +24,7 @@ tla2tools.jar:
 	(grep -q "\-\-algorithm" $*.tla && ${PLUSCAL} $*.tla) || true
 
 	@ echo "\n\n-------------- TLC ------------------"
-	${TLC} -workers ${WORKERS} -metadir metadir/check/$* $*.tla 2>&1 | tee $@.tmp
+	${TLC} -deadlock -workers ${WORKERS} -metadir metadir/check/$* $*.tla 2>&1 | tee $@.tmp
```
mullr, line 27: Is this enabling deadlock detection?

race-buffer/model/RaceBuffer.tla
```diff
@@ -6,111 +6,93 @@ CONSTANTS
...
-  /\ IF Read(i - 1).type = "PREFIX"
-     THEN Write(i, "SUFFIX")
-     ELSE \/ Write(i, "SINGLE")
-          \/ Write(i, "PREFIX")
-  /\ writes' = writes + 1
+  IF i = 0 
+  THEN WriteNewEntry(i) 
+  ELSE IF Read(i - 1).type = "PREFIX"
+       THEN Write(i, "SUFFIX")
+       ELSE WriteNewEntry(i)
```
mullr, line 55: Since what gets written depends on what is read, we should allow these to be schedule separately; I don't think we can consider this an atomic operation.

mullr: Or maybe it's ok, as this is essentially reprsenting state in the writer process?

NicholasSpringer: The Read(i) operator is just a shorthand to access the array. In practice, it is the user's responsibility to write the suffix they want, as there is no special treatment of suffixes from the writer side, except that a suffix is assumed to be right after a prefix when overwriting. Here I am just labeling the data as a suffix so the reader's output buffer can be checked for consistency.

race-buffer/model/RaceBuffer.tla
```diff
@@ -6,111 +6,93 @@ CONSTANTS
...
+
+(* Overwrite old entry by incrementing overwrite cursor *)
+OverwriteOldEntry == 
+    IF owcurs < BufCapacity 
+    THEN owcurs' = owcurs + 1
+    ELSE IF Read(owcurs).type = "PREFIX" 
+         THEN owcurs' = owcurs + 2
+         ELSE owcurs' = owcurs + 1
```
mullr, line 63: I think this is an actual read dependency

race-buffer/model/Makefile
NicholasSpringer: This is disabling deadlock detection

race-buffer/model/RaceBuffer.tla
NicholasSpringer: The writer is the only one that mutates writer state. It wouldn't make a difference if the reader were to read things before or after the `Read(owcurs).type = "PREFIX" ` check.

mullr: But, again, maybe ok in this case, since this is single writer?

NicholasSpringer: > But, again, maybe ok in this case, since this is single writer?

I believe so

NicholasSpringer: Yes, thinking about this more, the atomic steps that can be interspersed in any order are single writes to writer memory by the writer and single reads from writer memory by the reader.

PR #166, opened 2020-07-07
Use hashing to generate probe IDs

### LINKED ISSUE #150: Using hashing to generate probe and event ids
## Problem

Using offsets is painful as a usage pattern to try to make sure that probe & event IDs are unique. Additionally, the new "component" design depends on globally unique probe IDs, and so making attempts to ensure that with minimal user intervention in the common case is desirable.

## Considerations

We've previously made attempts to keep the symbol <-> ID mappings "stable", which means that they don't necessarily change every single time you run the code scanner. If you were to run the code scanner twice on the same source tree without any changes to your instrumentation call-sites, then the same IDs should be generated as were before. We believe we can still maintain this behavior, while also making probe/event IDs less likely to collide.

We can still provide the escape hatch of manually managing ID ranges should a user desire to do so.

## Proposal

The existing code scanner tools maintain their own heuristics for whether or not a call-site has "changed" enough to warrant issuing a new ID or notifying the user as to a potential bugaboo with reusing, duplicating, moving, etc. their call-sites around. We can leverage this behavior combined with hash-based id generation to get sparser allocation of IDs while maintaining their existing stability.

## Challenges

Hash output can't be the entire space of a `u32` because we reserve part of the range of IDs for ourselves, so we have to truncate and/or offset hash results to fit inside `TOTAL_SPACE - RESERVED_RANGE`. Just feeding the probe/event symbol names to the hash function doesn't do enough to avoid collisions in the event that two people happened to name something the same thing (e.g. `LOG_MESSAGE` event or something that might have a probable conflict) due to human inclination "inventing" the same thing in parallel in different contexts.

To resolve this issue we can consider using the scanner tools' knowledge of deciding when something has "changed" to be a trigger to generate something like a transient UUID and feed that UUID to the hash function. This improves the probability that we have collisions in the hash space based on statistical distribution rather than similarities in how humans name things. 
However, while leveraging the tools' "changed enough" heuristics and only doing this transient UUID -> hash -> ID keeps the probe/event IDs stable in the ways we care about, there is still the possibility for collision in two different contexts.

1. The first is at code scanning time, where different UUIDs hash to the same value, and the proposal for resolving this is, just generate another UUID and hash it, rinse repeat until no collision. The down side to this naive approach is that as the hash space fills up we'll probably do this a lot, which could be slow, but we'll be able to tell when we're having that problem, so we can resolve it in a future where we're dealing with systems that have ~2 Billion probes and/or ~2 Billion discrete event definitions.

1. Secondarily, we can have collisions on the probe IDs at integration time when multiple components are being brought together, and the proposal for resolving this is to allow the end-user to opt into defining ID ranges at the time they invoke the scanning tools to provide bounds for the ID range that can be output. The methodology that's used to constrain the `u32` hash space to be `TOTAL_SPACE - RESERVED_RANGE` can probably be applied somewhat similarly here to make sure the hash output stays within some supplied bounds.

jonlamb-gh:
Based on #161
Closes #150


modality-probe-cli/src/manifest_gen/id_gen.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    fn token_hash(&self, token: &str) -> u32 {
+        let mut hasher = Sha3_256::new();
+        hasher.update(self.uuid.as_bytes());
+        hasher.update(token.as_bytes());
+        let hash = hasher.finalize();
+        let bytes: &[u8; 32] = hash.as_ref();
+        let be16_bytes = u16::from_be_bytes([bytes[0], bytes[1]]);
+        let be32_bytes = u32::from_be_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]);
```
pittma, line 73: What do you think about
```rust
        let be16_bytes = u16::from_be_bytes(bytes[0..2].try_into().expect("getting two bytes"));
        let be32_bytes = u32::from_be_bytes(bytes[0..4].try_into().expect("getting four bytes"));
```
?

modality-probe-cli/src/manifest_gen/id_gen.rs
```diff
@@ -0,0 +1,207 @@
...
+        let id = u32::from(be16_bytes)
+            .overflowing_mul(0xFFFF_FFFF)
+            .0
+            .overflowing_add(be32_bytes)
+            .0;
+        id % self.id_range.inclusive_end.get()
+    }
+}
+
+impl NonZeroIdRange {
```
pittma, line 83: Want to include this in the NonZeroIdRange impl block above?

modality-probe-cli/src/manifest_gen/invocations.rs
```diff
@@ -513,21 +526,28 @@ impl Invocations {
...
+        loop {
+            let next_id = gen.hashed_id(probe_name);
 
-        let next_id = components
-            .iter()
-            .map(|c| c.probes.next_available_probe_id())
-            .max()
-            .unwrap_or(1);
+            let id_already_taken = existing_probe_ids.iter().any(|&id| id == next_id.get());
+            let is_valid_probe_id = modality_probe::ProbeId::new(next_id.get()).is_some();
```
pittma, line 539: I tried to convert this to an `if let ...`
```rust
            let id_already_taken = existing_probe_ids.iter().any(|&id| id == next_id.get());
            if let Some(probe_id) = modality_probe::ProbeId::new(next_id.get()) {
                if !id_already_taken {
                    existing_probe_ids.push(next_id.get());
                    return probe_id;
                }
            }
```
 but couldn't and was surprised by the reason:

```
error[E0308]: mismatched types
   --> modality-probe-cli/src/manifest_gen/invocations.rs:542:28
    |
542 |                     return probe_id;
    |                            ^^^^^^^^ expected struct `probes::ProbeId`, found struct `modality_probe::id::ProbeId`
```
Can the CLI use the core lib's definition of probe id instead?

jonlamb-gh: Yeah I think we can get things switched over now that we've got the std-feature-gated serde impls.
Mind if we make a separate ticket for it?

modality-probe-cli/src/manifest_gen/id_gen.rs
jonlamb-gh: Sounds good, c2ba416587d310f0642db812259c0e3fd480c893

modality-probe-cli/src/manifest_gen/id_gen.rs
jonlamb-gh: :+1: c2ba416587d310f0642db812259c0e3fd480c893

PR #169, opened 2020-07-08
Add produce/merge snapshot bytes API

jonlamb-gh:
* Rename `distribute_snapshot` to `produce_snapshot`
* Add `produce_snapshot_bytes()` and `merge_snapshot_bytes()`


modality-probe-capi/modality-probe-capi-impl/src/lib.rs
```diff
@@ -199,29 +196,55 @@ fn distribute_error_to_modality_probe_error(
...
+    history_destination: *mut u8,
+    history_destination_bytes: usize,
+    out_written_bytes: *mut usize,
+) -> ModalityProbeError {
+    let probe = match probe.as_mut() {
+        Some(t) => t,
+        None => return MODALITY_PROBE_ERROR_NULL_POINTER,
+    };
+    match probe.produce_snapshot_bytes(core::slice::from_raw_parts_mut(
+        history_destination,
```
ZackPierce, line 233: Should we also null-check the destination pointer?

jonlamb-gh: :+1: 228c86cb03e059e38c0e19ffc480bd5884a1fe35

PR #170, opened 2020-07-09
add racebuffer reporting design

NicholasSpringer:
Proposed method of creating reports from RaceBuffer-stored compact log


NicholasSpringer:
After thinking about this more, I'm not sure if the `is_next_suffix` state is necessary. Instead of splitting double-item entries between chunks, both items can be put into the later chunk. It doesn't seem like this would be too much of a detriment for performance, and it improves simplicity.

PR #171, opened 2020-07-13
integrate components as a first-class thing for export subcommand

### LINKED ISSUE #172: Export command should use components rather than discrete events and probes csv files
Follow-on work from https://github.com/auxoncorp/modality-probe/issues/155.

pittma:
Closes #172.

### DIFF
--- a/modality-probe-cli/src/export.rs
+++ b/modality-probe-cli/src/export.rs
@@ -5,7 +5,10 @@ use std::{collections::HashMap, path::PathBuf, str::FromStr};
 
 use structopt::StructOpt;
 
-use crate::graph::{self, digraph::Digraph, meta::*, Cfg};
+use crate::{
+    component::Component,
+    graph::{self, digraph::Digraph, meta::*, Cfg},
+};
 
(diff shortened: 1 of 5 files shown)


modality-probe-cli/src/export.rs
```diff
@@ -5,24 +5,25 @@ use std::{collections::HashMap, path::PathBuf, str::FromStr};
...
     /// eliding the events inbetween.
-    #[structopt(long)]
+    #[structopt(short, long)]
     pub interactions_only: bool,
-    /// The path the probes.csv for a component.
-    #[structopt(long)]
-    pub probes: PathBuf,
+    /// The path to a component directory. To include multiple
+    /// components, provide this switch multiple times.
+    #[structopt(short, long)]
```
jonlamb-gh, line 23: Maybe worth adding `required = true` to have the cli complain about missing `--components`?

pittma: Aye. https://github.com/auxoncorp/modality-probe/pull/171/commits/21a569afdff6c44ef0ef9a724cc197a310e5093a

PR #173, opened 2020-07-14
Include the interacting clocks whether they're stale or not

pittma:


pittma:
This'll be included in #179.

PR #177, opened 2020-07-16
Split probe logical clock into epoch/clock

mullr:


src/history.rs
```diff
@@ -47,6 +44,9 @@ pub const MIN_HISTORY_SIZE_BYTES: usize = size_of::<DynamicHistory>()
     + MIN_CLOCKS_LEN * size_of::<LogicalClock>()
     + MIN_LOG_LEN * size_of::<CompactLogItem>();
 
+const EPOCH_WRAPAROUND_THRESHOLD_TOP: u16 = ProbeEpoch::MAX - 3;
```
pittma, line 47: This offset is to order in the face of a rollover / restart I assume? Is this documented somewhere? If not, would you mind writing  comment here explaining what these are for?

mullr: This is for handling epoch rollover. Will link you to the doc out of band.

collectors/modality-probe-udp-collector/src/lib.rs
```diff
@@ -204,14 +204,30 @@ mod tests {
...
-        LogicalClock, Probe,
+        compact_log::LogEvent, report::wire::OwnedLogSegment, BulkReporter, CausalSnapshot,
+        ChunkedReporter, LogicalClock, Probe, ProbeId,
     };
 
     use modality_probe_collector_common::*;
 
     use super::*;
 
+    struct DefaultableCausalSnapshot(CausalSnapshot);
```
jonlamb-gh, line 215: Doesn't look like we're using this struct anymore, is it left in intentionally?

src/wire/causal_snapshot.rs
```diff
@@ -111,11 +118,18 @@ impl<T: AsRef<[u8]>> WireCausalSnapshot<T> {
         }
     }
 
-    /// Return the `count` field
+    /// Return the `epoch` field
     #[inline]
-    pub fn count(&self) -> u32 {
+    pub fn probe_epoch(&self) -> u16 {
```
jonlamb-gh, line 123: Looks like the getter is returning a `u16`, but the setter is taking a `ProbeEpoch`.
Can they both use `ProbeEpoch`?

src/wire/causal_snapshot.rs
```diff
@@ -111,11 +118,18 @@ impl<T: AsRef<[u8]>> WireCausalSnapshot<T> {
...
         let data = self.buffer.as_ref();
-        le_bytes::read_u32(&data[field::COUNT])
+        le_bytes::read_u16(&data[field::PROBE_EPOCH])
+    }
+
+    /// Return the `ticks` field
+    #[inline]
+    pub fn probe_ticks(&self) -> u16 {
```
jonlamb-gh, line 130: Based on the setter method names, seems like this should be named `ticks()`.
Same with `probe_epoch()`, ought to be `epoch()`.

collectors/modality-probe-udp-collector/src/lib.rs
mullr: I accidentally brought this back during a rebase; will remove.

PR #185, opened 2020-07-31
Unifying the Reporting API

### LINKED ISSUE #179: Unifying the Reporting API
With a few structural changes, the chunked reporting API can subsume the bulk reporting API with something that looks more like streaming the log directly from the probe.

### Reporting API
The reporting API will retain what is now the bulk API.

```c
/*
 * Write a report of recorded events and logical clock
 * data to a supplied destination.
 *
 * Populates the number of bytes written in out_written_bytes.
 */
size_t modality_probe_report(
    modality_probe *probe,
    uint8_t *log_report_destination,
    size_t log_report_destination_bytes,
    size_t *out_written_bytes);
```

This function will take the given buffer and first include the wire header, then the frontier clocks, and then finally as much of the event log as possible.

* **Error case**: If the buffer is too small for any events to fit, that is, there’s room for only the header and some or all of the clocks, then a “special report” is created containing the header, then a single event stating that the report buffer is too small.
* **Error case**: If the buffer is too small for even the header to fit, it is a hard error.

The API for chunked reporting will be removed.

#### Log Storage
The log storage can be moved from the `FixedSliceVec`-backed storage to one that uses `RaceBuffer`<sup>1</sup>. This allows for invocations of the reporting API to simply move the read cursor as it sees fit.

#### Frontier Clocks
The frontier clocks that the probe keeps—those clocks that are the maximum clock tick value that that probe has seen from any neighbors—will be made distinct from trace clocks, those clocks that signify an interaction in the SUA and are a part of the trace.

#### Report Wire Format
The wire format will be a variation on the chunked reporting format:

<table>
  <tr>
    <td>Fingerprint <br> (4 bytes)</td>
    <td>Probe Id <br> (4 bytes)</td>
    <td>Local Clock <br> (4 bytes)</td>
    <td>Report<br>Sequence<br>Number <br> (8 bytes)</td>
    <td>Num Clocks <br> (2 bytes)</td>
    <td>Num Log Items <br> (4 bytes)</td>
    <td>Payload<sup>2</sup> <br> (N bytes)</td>
  </tr>
</table>

With the addition of a sequence number that resets when the local clock is incremented. This is to order reports that are drawn from the probe between a tick to its local clock.

##### Fingerprint
Reports will use the new fingerprint `0x4D525054 (MRPT)`.

#### Log Storage
Swap out `FixedSliceVec` for `RaceBuffer` as discussed in a previous section.

#### Report Sequence Number
The probe will need the addition of a 16-bit counter that is ticked when the report API is invoked, this
counter wraps around when it overflows.

#### Chunked Reporting State
This will be removed. The probe needn’t track that a report is in progress any longer.

---

1. Note that @NicholasSpringer has done most of this work already [in a feature branch](https://github.com/auxoncorp/modality-probe/tree/race-buffer-log-storage).
2. Frontier clocks first, then log data.

jonlamb-gh:
Closes #179


collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -37,78 +42,52 @@ newtype! {
...
 
-impl From<(ProbeId, ProbeEpoch, ProbeTicks)> for LogEntryData {
-    fn from((id, epoch, clock): (ProbeId, ProbeEpoch, ProbeTicks)) -> LogEntryData {
-        LogEntryData::LogicalClock(id, epoch, clock)
-    }
+#[derive(Debug, PartialEq)]
+pub struct Report {
+    pub probe_id: ProbeId,
+    pub probe_clock: LogicalClock,
+    pub seq_num: u64,
```
ZackPierce, line 63: We have a `SequenceNumber` type hanging out nearby, should we be using it here?

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
-                    log_entries_buffer.push(LogEntry {
-                        session_id,
-                        segment_id: next_segment_id.into(),
-                        segment_index,
-                        probe_id,
-                        data: LogEntryData::Event(*ev),
-                        receive_time,
-                    });
+        let mut interpret_next_as = Next::DontKnow;
+        for u32_bytes in payload[clocks_len..].chunks_exact(mem::size_of::<LogEntry>()) {
```
ZackPierce, line 426: This interpretation code probably has a canonical place to live somewhere in our log module (or, if there's divergence now, in the reporting module).  In the previous version of this I wrote some iterators in this vein to prevent rework across collector implementations, and would recommend the same here, since it seems probably unnecessary to require collectors to know the deep details of the log-format.

ZackPierce: Now that I look up and see that this is in a cross-collector lib crate, the above comment seems mostly superfluous.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
+        };
+
+        let clocks_len = report.n_clocks() as usize * mem::size_of::<LogicalClock>();
+        let mut probe_id = None;
+        let payload = report.payload();
+        for u32_bytes in payload[..clocks_len].chunks_exact(mem::size_of::<LogEntry>()) {
+            let raw = le_bytes::read_u32(u32_bytes);
+            if probe_id.is_none() {
+                probe_id = ProbeId::new(
+                    unsafe { LogEntry::new_unchecked(raw) }.interpret_as_logical_clock_probe_id(),
```
ZackPierce, line 412: Why are we skipping the checks here?

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -37,78 +42,52 @@ newtype! {
...
 
-/// The data that may be attached to a log entry
-#[derive(Debug, Eq, PartialEq, Clone)]
-pub enum LogEntryData {
-    Event(EventId),
-    EventWithPayload(EventId, u32),
-    LogicalClock(ProbeId, ProbeEpoch, ProbeTicks),
-}
+#[derive(Debug)]
+pub struct SerializationError;
```
ZackPierce, line 51: This seems like it would be a maddening error to receive, since there's no context whatsoever and greater than 5 ways it might be triggered.

collectors/modality-probe-collector-common/src/lib.rs
ZackPierce: Since we're interpreting raw bytes, this seems like exactly the spot to do the range checks.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
             )
     }
 
-    prop_compose! {
-        pub(crate) fn gen_raw_user_event_id()(raw_id in 1..=EventId::MAX_USER_ID) -> u32 {
-            raw_id
-        }
-    }
+    #[test]
+    fn report_e2e() {
```
ZackPierce, line 651: Could we please also get a proptest around reporting round-tripping here?  I'm talking about generated Report -> &[u8] -> Report equality checking.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
+        let mut wire = WireReport::new_unchecked(bytes);
+        wire.check_len()?;
+        wire.set_fingerprint();
+        wire.set_probe_id(self.probe_id);
+        wire.set_clock(modality_probe::pack_clock_word(
+            self.probe_clock.epoch,
+            self.probe_clock.ticks,
+        ));
+        wire.set_seq_num(self.seq_num);
+        wire.set_n_clocks(self.frontier_clocks.len() as _);
```
ZackPierce, line 486: Seems like we should return an error rather than panicking if the `len()` usize does not fit into the wire's primitive size.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
+            .map(|e| match e {
+                EventLogEntry::Event(_) => 1,
+                EventLogEntry::EventWithPayload(_, _) => 2,
+                EventLogEntry::TraceClock(_) => {
+                    mem::size_of::<LogicalClock>() / mem::size_of::<u32>()
+                }
+            })
+            .sum();
+
+        wire.set_n_log_entries(num_u32_entries as _);
```
ZackPierce, line 500: Same concern re: `as` casting causing panics rather than returning errors.

collectors/modality-probe-collector-common/src/lib.rs
pittma: Oh, for sure. There was once a TODO here to fill this in and add error derivation. It must have gotten eaten in the shuffle.

mullr: I fixed this locally; will push it shortly.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +354,414 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
-            });
+enum Next {
+    Clock(ProbeId),
+    Payload(EventId),
+    DontKnow,
+}
 
-            segment_index += 1;
+impl Report {
+    pub fn write_into_le_bytes(&self, bytes: &mut [u8]) -> Result<usize, ReportWireError> {
```
pittma, line 476: Is this expected to be used outside of tests?

collectors/modality-probe-collector-common/src/lib.rs
pittma: For sure. I think what's missing here is two things:
1. first that the `LogEntry` had the clock bit set and _then_ interpreting it.
2. `ok_or_else`-ing the `ProbeId::new` call.

collectors/modality-probe-collector-common/src/lib.rs
jonlamb-gh: c0adc77eab6c64507ba5d6b58460c064b1aa671d

collectors/modality-probe-collector-common/src/lib.rs
jonlamb-gh: I think I covered this in c0adc77eab6c64507ba5d6b58460c064b1aa671d

collectors/modality-probe-collector-common/src/lib.rs
jonlamb-gh: :+1: c0adc77eab6c64507ba5d6b58460c064b1aa671d

collectors/modality-probe-collector-common/src/lib.rs
jonlamb-gh: :+1: c0adc77eab6c64507ba5d6b58460c064b1aa671d

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -283,320 +370,498 @@ impl TryFrom<&LogFileRow> for LogEntry {
...
-                ),
-                receive_time,
-            });
+impl Report {
+    pub fn write_into_le_bytes(&self, bytes: &mut [u8]) -> Result<usize, SerializationError> {
+        if self.frontier_clocks.len() > std::u16::MAX as usize {
+            return Err(SerializationError::TooManyFrontierClocks(
+                self.frontier_clocks.len(),
+            ));
+        } else if self.event_log.len() > std::u32::MAX as usize {
```
ZackPierce, line 502: IIRC, the value used later in `set_n_log_entries` may actually be larger than this since more than one log entry may be used for, say, events with payloads. Thus a simple len check here is not sufficient.  Probably should do the check against the computed `num_u32_entries` below.

jonlamb-gh: Ah yeah, good catch
507ba4c9ab2b35a409247c8ced1e51303cbe5ab0

race-buffer/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+}
+
+pub mod writer;
+
+#[cfg(feature = "std")]
+pub mod reader;
+
+#[cfg(feature = "std")]
+#[cfg(test)]
+pub mod tests {
```
ZackPierce, line 30: Could I request some tests that capture the exact expected behavior when we write more than the buffer can store? In that same vein, tests that show what happens when we attempt to read when the buffer is empty? Not-quite-equivalently, what we expect to happen when we are both reading and writing but doing one of those more often than the other.  IIUC, said tests wouldn't necessarily have to be multithreaded to be useful codifications.

jonlamb-gh: cc @NicholasSpringer ^^

collectors/modality-probe-collector-common/src/lib.rs
ZackPierce: I would expect this to be useful if at some point someone wants to use this `Report` structure to synthesize a report outside of the context of the provided probe implementation.  For now, `#[cfg(test)]` is probably appropriate, though.

race-buffer/src/reader.rs
```diff
@@ -0,0 +1,139 @@
...
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+/// Used to "snap" observations of the RaceBuffer's write cursor
+/// or entries in the RaceBuffer's backing storage. Each Snapper access
+/// should have sequential consistency with the memory access of the RaceBuffer writer
+pub trait Snapper<E>
+where
+    E: Entry,
+{
+    /// Take a snapshot of the write cursor of the RaceBuffer
+    fn snap_wcurs(&self) -> Result<usize>;
```
ZackPierce, line 15: Since this is a library (rather than a binary), I recommend using precise errors where possible in order to allow targeted remediation without requiring downcasting. Here the most direct solution is to add an associated `type Error: std::error::Error` to the `Snapper` trait, and have these methods return `Result<MethodSpecificType, Self::Error>`.

race-buffer/src/reader.rs
```diff
@@ -0,0 +1,139 @@
...
+            snapper,
+            rcurs: 0,
+            storage_cap,
+            stored_prefix: None,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Attempt to read all new entries in race buffer into given read buffer
+    pub fn read(&mut self, rbuf: &mut Vec<Option<E>>) -> Result<()> {
```
ZackPierce, line 56: A possible API adjustment (not critical) that would save some space and maybe be a bit clearer to interpret:

Rather than handling possibly-missing items in the out-buffer with `Option`s like `vec![None, None, Some(e), Some(e)]`, how about instead storing the direct entries and returning the number missed?

E.G. `fn read(&mut self, rbuf: &mut Vec<E>) -> Result<usize, S::Error>`

This could save users of the library downstream Option-unwrapping (and, less importantly, reduce the memory footprint of the needed rbuf).

race-buffer/src/reader.rs
```diff
@@ -0,0 +1,139 @@
...
+        if pre_owcurs > self.rcurs + self.storage_cap {
+            // If writer has overwritten unread entries between reads, store nils and correct rcurs
+            let num_missed = pre_owcurs - (self.rcurs + self.storage_cap);
+            self.store_missed_items(num_missed, rbuf);
+            self.rcurs = pre_owcurs - self.storage_cap;
+        }
+
+        // Perform reads into snapshot buffer up to write cursor
+        let first_read = self.rcurs;
+        let mut buf_snapshot: Vec<E> = Vec::new();
```
ZackPierce, line 69: I'm unclear on why we need this intermediate buffer. Can we write directly to `rbuf` (and manage any necessary initial-index-offsetting needed)?

race-buffer/src/reader.rs
```diff
@@ -0,0 +1,139 @@
...
+        for entry in &mut buf_snapshot[overlap..] {
+            self.store(*entry, rbuf);
+        }
+        Ok(())
+    }
+
+    /// Store given entry in given read buffer
+    #[inline]
+    fn store(&mut self, entry: E, rbuf: &mut Vec<Option<E>>) {
+        match self.stored_prefix {
```
ZackPierce, line 101: Unessential tip: Consider using `Option::take` if you have a pattern of trying to use a stored optional value that you intend to replace or clear-out as a result of the operations. E.G.

```rust
if let Some(prefix) = self.stored_prefix.take() {
    rbuf.push(Some(prefix));
    rbuf.push(Some(entry));
} else if entry.is_prefix() {
    self.stored_prefix = Some(entry);
} else {
    rbuf.push(Some(entry));
}
```

race-buffer/src/lib.rs
NicholasSpringer: @ZackPierce Just so you know, there is another branch off of `reporty-unity` called `race-buffer-modifications` with quite a few changes to the race buffer. It has some more tests and other modifications, but the things you have suggested so far still need to be addressed.

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,252 @@
+use crate::{get_cursor_index, Entry};
+use core::iter::Iterator;
+use core::mem::MaybeUninit;
+
+#[cfg(not(feature = "std"))]
+use core::fmt;
+#[cfg(feature = "std")]
```
ZackPierce, line 7: It's fine to just use `core::fmt` even when the standard library is available.  Can remove some configuration flagging in this vein.

src/payload.rs
```diff
@@ -0,0 +1,516 @@
...
+                return (num_written, num_read, clocks_overflowed);
+            }
+        };
+
+        if cur_entry.has_clock_bit_set() {
+            if output.free_capacity() < 2 {
+                // Could not fit clock
+                return (num_read, num_written, clocks_overflowed);
+            }
+            // RaceBuffer guarantees that if prefix is iterated over/read, then suffix will also be available and present
```
ZackPierce, line 174: If this is the case, then we should consider whether the `log_iter` provided to this function should have a different `Item` type that reflects this promise -- e.g. some sort of `enum` that yields either a single non-prefix value or a prefix/suffix pair.  

This may be too much type-attention to ask at this point, but may be a useful thought experiment for a possible future improvement.

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,252 @@
...
+                && unsafe { self.read_storage(self.wcurs).is_prefix() }
+            {
+                // If overwriting double, increment overwrite cursor by 2
+                overwritten = unsafe {
+                    OverwrittenEntry::Double(
+                        self.read_storage(self.owcurs),
+                        self.read_storage(self.owcurs + 1),
+                    )
+                };
+                self.owcurs += 2;
```
ZackPierce, line 136: Am I correct in assuming the intent is that `owcurs` and `wcurs` are supposed to wrap when they hit `usize::MAX`?  

If so, we'll need to modify this and the other places the values are updated to something like `self.owcurs = self.owcurs.wrapping_add(2)`.  If special remediation is required at the point of wrap-around, then https://doc.rust-lang.org/std/primitive.usize.html#method.overflowing_add can be used to detect when that happens.

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+
+    #[inline]
+    pub fn iter<'b>(&'b self, start_rcurs: usize) -> RaceBufferIter<'a, 'b, E> {
+        RaceBufferIter::new(self, start_rcurs)
+    }
+
+    #[inline]
+    pub fn as_slice(&self) -> &[MaybeUninit<E>] {
```
ZackPierce, line 194: Seems odd to expose these internals as a public function.  Can we avoid this?

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    #[inline]
+    fn write_to_storage(&mut self, curs: usize, entry: E) {
+        self.storage[get_cursor_index(self.storage.len(), curs, self.use_base_2_indexing)] =
+            MaybeUninit::new(entry);
+    }
+
+    #[inline]
+    pub fn read(&self, curs: usize) -> ReadEntry<E> {
```
ZackPierce, line 172: I don't see any tests that demonstrate the interaction of `RaceBuffer::read` or `RaceBuffer::iter` with `RaceBuffer::write` and `RaceBuffer::write_double` (the tests in lib.rs presently explore the remote-reader-style patterns). Covering the basics seems relevant here.

race-buffer/src/reader.rs
NicholasSpringer: This is possible, but it would require copying the entries anyway if any items were missed during the read.

src/payload.rs
NicholasSpringer: Yeah, that sounds reasonable to me.

race-buffer/src/writer.rs
NicholasSpringer: This is dealt with in the `race-buffer-modifications` branch

race-buffer/src/writer.rs
NicholasSpringer: Right now the `as_slice` function is used to get a pointer to the storage to ensure it isn't overlapping with the clocks list when creating a new DynamicHistory. Other than that, it isn't necessary

race-buffer/src/writer.rs
NicholasSpringer: This is present on `race-buffer-modifications`

race-buffer/src/reader.rs
ZackPierce: Reducing the number of involved entities is probably a general win for code understandability. Removing the intermediate Vec also cuts out some allocation, too.  Code clarity is the higher priority, and since it's pretty subjective, so I wouldn't call this quibble essential.

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,498 @@
...
+            "Storage capacity must be at least {} but no greater than {}.",
+            MIN_STORAGE_CAP,
+            SEQN_MOD_MAX / 2,
+        )
+    }
+}
+
+#[inline]
+/// Round given length down to a power of 2
+fn round_to_power_2(length: usize) -> usize {
```
jonlamb-gh, line 36: @NicholasSpringer Are there any test for this?

race-buffer/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "race-buffer"
+version = "0.1.0"
+authors = ["Nicholas Springer <nspringer@auxon.io>"]
+edition = "2018"
+
+[dependencies]
+crossbeam = {version = "0.7.3", optional = true}
```
ZackPierce, line 8: Shouldn't `crossbeam` and `rand` be in `dev-dependencies` not regular dependencies?

src/error.rs
```diff
@@ -4,20 +4,43 @@
...
 ///
 ///
 /// probe ids must be greater than 0 and less than ProbeId::MAX_ID
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct InvalidProbeId;
 
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidProbeId {}
+
+#[cfg(feature = "std")]
```
ZackPierce, line 37: The `Display` trait is available in `core`, should be able to implement without feature-flagging.

src/error.rs
```diff
@@ -29,6 +52,32 @@ pub enum InitializationError {
...
+impl std::error::Error for InitializationError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            InitializationError::InvalidProbeId => None,
+            InitializationError::StorageSetupError(e) => Some(e),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
```
ZackPierce, line 65: Same, no need to feature-flag this out, I think.

src/error.rs
```diff
@@ -46,6 +95,24 @@ pub enum StorageSetupError {
     NullDestination,
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for StorageSetupError {}
+
+#[cfg(feature = "std")]
+impl Display for StorageSetupError {
```
ZackPierce, line 102: Ditto re: unnecessarily flagged-out bit of usefulness

ZackPierce: I'm going to stop marking this pattern in GH and leave its find-and-cleanup as an exercise to the implementer.

src/history.rs
```diff
@@ -1,72 +1,72 @@
...
+);
+
+#[cfg(not(target_pointer_width = "32"))]
+const_assert_eq!(
+    4 + size_of::<ProbeId>()
+        + size_of::<u32>()
+        + size_of::<FixedSliceVec<'_, LogicalClock>>()
+        + size_of::<LogicalClock>()
+        + size_of::<RaceLog<'_>>()
+        + size_of::<u64>()
```
ZackPierce, line 68: It looks like the difference between these two is a pointer width kind of difference (usize when target_pointer_width = 32, u64 otherwise [and otherwise probably means 64-bit]), so I'm curious why using size_of::<usize> for this part of the equation doesn't work for both scenarios.

src/history.rs
```diff
@@ -155,126 +153,184 @@ impl<'a> DynamicHistory<'a> {
...
+    #[inline]
+    fn write_to_log(&mut self, entry: LogEntry) {
+        let overwritten = self.log.write(entry);
+        self.merge_overwritten_clock(overwritten);
+    }
+
+    #[inline]
+    fn write_double_to_log(&mut self, prefix: LogEntry, suffix: LogEntry) {
+        let (first_overwritten, second_overwritten) = self.log.write_double(prefix, suffix);
+        self.merge_overwritten_clock(first_overwritten);
```
ZackPierce, line 212: These entries might not be clocks, and we should know this at the call site.  I would rather we have `write_event_with_payload_to_log` and `write_clock_pair_to_log` than tossing both pair-types to a method that's named for just clocks (and conducts a redundant check inside).

race-buffer/src/writer.rs
```diff
@@ -0,0 +1,501 @@
...
+        if num_missed == 0 && read_seqn != self.write_seqn {
+            PossiblyMissed::Entry(unsafe { self.read_storage(read_seqn) })
+        } else {
+            PossiblyMissed::Missed
+        }
+    }
+
+    /// Iterate over the entries, where None represents a missed entry
+    #[inline]
+    pub fn iter<'b>(&'b self, start_seqn: u32) -> RaceBufferIter<'a, 'b, E> {
```
ZackPierce, line 219: It feels strange to have to supply a starting index to an iterator, when for the probe use case we always need to start at the "start."  

I get that for fully separated reader/writer pairs we track the cursors distinctly and this API makes sense in that context, but perhaps there's a slightly higher level abstraction we could provide around RaceBuffer that manages a read cursor to give simplified access to reading from the head, deleting from the head, and writing to the tail.

src/history.rs
```diff
@@ -155,126 +153,184 @@ impl<'a> DynamicHistory<'a> {
...
+        }
+
+        let self_clock = self.self_clock;
+        let mut report = WireReport::new_unchecked(destination);
+        report.set_fingerprint();
+        report.set_probe_id(self.probe_id);
+        report.set_clock(crate::pack_clock_word(self_clock.epoch, self_clock.ticks));
+
+        // We can't store at least the frontier clocks and a single two-word item
+        if report.as_ref().len() < WireReport::<&[u8]>::buffer_len(self.clocks.len(), 2) {
```
ZackPierce, line 270: In this first branch, I think we need only one word-sized item, the event id of the "not enough space" event.

ZackPierce: A very similar check at the top returns an error. I think we should probably be consistent and do the same.

I understand it's a trade-off between being able to produce a report that contains an error-signaling-event (which might make it to the collector for analysis) versus producing a local error, and in this case I think the local error is preferable because the remediation is certainly going to be locally-specific, and also not particularly difficult.

src/history.rs
```diff
@@ -155,126 +153,184 @@ impl<'a> DynamicHistory<'a> {
...
+    #[inline]
+    pub(crate) fn produce_snapshot(&mut self) -> Result<CausalSnapshot, ProduceError> {
+        let snap = CausalSnapshot {
+            clock: self.self_clock,
             reserved_0: 0,
             reserved_1: 0,
-        })
+        };
+        self.increment_local_clock();
+        self.write_clocks_to_log(&[self.self_clock]);
```
ZackPierce, line 316: I think now that we only ever write one clock to the log at a time, I think we ought to transform this fn to `write_clock_to_log`.

src/history.rs
```diff
@@ -79,15 +79,13 @@ pub struct DynamicHistory<'a> {
     /// The number of events seen since the current
     /// probe's logical clock last increased.
     pub(crate) event_count: u32,
-    pub(crate) chunked_report_state: ChunkedReportState,
     /// Invariants:
     ///   * The first clock is always that of the local probe id
     pub(crate) clocks: FixedSliceVec<'a, LogicalClock>,
```
ZackPierce, line 84: This field is now the frontier clocks known as of the read-for-reporting head of the log, right?

src/history.rs
```diff
@@ -155,126 +153,184 @@ impl<'a> DynamicHistory<'a> {
...
+
+    #[inline]
+    fn merge_overwritten_clock(&mut self, overwritten: OverwrittenEntry<LogEntry>) {
+        if let OverwrittenEntry::Double(one, two) = overwritten {
+            // If what we get out of the log is a clock, merge it into clocks list
+            if one.has_clock_bit_set() {
+                if let Some(id) = ProbeId::new(one.interpret_as_logical_clock_probe_id()) {
+                    let (epoch, ticks) = crate::unpack_clock_word(two.raw());
+                    self.merge_clock(LogicalClock { id, epoch, ticks });
+                }
```
ZackPierce, line 198: I wonder if there's anything we should do in the `else` branch. Log an error-ish event?

src/log.rs
```diff
@@ -0,0 +1,154 @@
...
+
+impl core::fmt::Debug for LogEntry {
+    #[inline]
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        f.write_fmt(format_args!("LogEntry({})", self.0))
+    }
+}
+
+impl Entry for LogEntry {
+    fn is_prefix(&self) -> bool {
```
ZackPierce, line 112: This function might still return a misleading value if it is used on a payload / clock value.  Maybe add some warning comment?

src/payload.rs
```diff
@@ -0,0 +1,516 @@
...
+                    epoch: ProbeEpoch(0),
+                })
+                .1,
+            ),
+        ];
+
+        let mut payload = [0u8; 128];
+        let (num_written, num_read, did_clocks_overflow) = write_report_payload(
+            log.iter().copied(),
+            &mut FrontierClocksFSV::new(&mut frontier_clocks),
```
ZackPierce, line 295: Why are we making a new instance here rather than holding on to the instance created earlier in the test?

src/payload.rs
```diff
@@ -0,0 +1,516 @@
...
+                id: probe_id,
+                ticks: ProbeTicks(1),
+                epoch: ProbeEpoch(0),
+            }],
+        )
+    }
+
+    /// No space for internal event, don't write anything
+    #[test]
+    fn no_space_for_missed() {
```
ZackPierce, line 447: Could we also get a test for the other semi-error case, when there wasn't enough room in the frontier clock vec-like object to add a merged clock to it and subsequently a warning event was recorded in the log?

src/wire/report.rs
```diff
@@ -0,0 +1,412 @@
...
+        r.set_n_clocks(2);
+        r.set_n_log_entries(3);
+        r.payload_mut().copy_from_slice(&PAYLOAD_BYTES[..]);
+        assert_eq!(r.check_fingerprint(), Ok(()));
+        assert_eq!(r.check_payload_len(), Ok(()));
+        assert_eq!(&r.into_inner()[..], &MSG_BYTES[..]);
+    }
+
+    #[test]
+    fn deconstruct() {
```
ZackPierce, line 351: Could we have an API-oriented struct-independent round-tripping test? One where we wrap a byte array with WireReport, set the field values, then hand over the byte array to a fresh WireReport instance and confirm that the newly-accessible values match what was originally written? This would be a backstop against accidental statefulness of the WireReport instance and how that might affect interpretability.

src/history.rs
ZackPierce: Seems to be consistently handled thus far.

PR #188, opened 2020-08-05
Unifying the Reporting API: without RaceBuffer for now

### LINKED ISSUE #179: Unifying the Reporting API

jonlamb-gh:
Closes #179 

Based on #185 , but replaces the RaceBuffer with a basic ring buffer to get us by until the RaceBuffer work settles.


src/history.rs
```diff
@@ -155,126 +139,215 @@ impl<'a> DynamicHistory<'a> {
...
+            report.set_n_clocks(0);
+            report.set_n_log_entries(1);
+            let payload = report.payload_mut();
+            payload[..size_of::<LogEntry>()].copy_from_slice(
+                &EventId::EVENT_INSUFFICIENT_REPORT_BUFFER_SIZE
+                    .get_raw()
+                    .to_le_bytes(),
+            );
+        } else {
+            if self.log_items_missed != 0 {
```
jonlamb-gh, line 252: I just realized that I'm not actually incrementing this anywhere. Will fix.

jonlamb-gh: Fixed in 5ef04f4

src/history.rs
```diff
@@ -155,126 +139,231 @@ impl<'a> DynamicHistory<'a> {
...
+                self.record_event_with_payload(
+                    EventId::EVENT_LOG_ITEMS_MISSED,
```
ZackPierce, line 261: Should write this into the report log output (after the frontier clocks, before the rest of the log), but not into the probe's self.log.

jonlamb-gh: Good catch, fixed in cc1273e

tests/integration_tests.rs
```diff
@@ -264,3 +265,59 @@ fn report_missed_log_items() -> Result<(), ModalityProbeError> {
...
+    fn reports_never_end_between_multi_log_items(
+        num_events in 0_usize..256_usize,
+        num_events_with_payload in 0_usize..256_usize,
+        report_buffer_space in 0_usize..256_usize,
+    ) {
+        let mut storage = [0u8; 512];
+        let probe = ModalityProbe::try_initialize_at(&mut storage, 1).unwrap();
+        let event_a = EventId::new(2).unwrap();
+        let event_b = EventId::new(3).unwrap();
+        let event_b_payload = 4;
```
ZackPierce, line 280: I think this would be a stronger test if the payload was allowed to vary freely. This would help cover the cases where the payload might, incidentally, have its "clock bit" set or its "event-with-payload bit" set and we have to rely on the correctness of the log-management to avoid misinterpretation.

ZackPierce:
:+1:

PR #190, opened 2020-08-06
Add a log entry peek pattern

jonlamb-gh:
Add a log entry peek pattern to avoid dropping log entries when bailing out of a report early.

### DIFF
--- a/src/ring.rs
+++ b/src/ring.rs
@@ -140,6 +140,17 @@ impl<'a> LogEntryRingBuffer<'a> {
         }
     }
 
+    /// Read a single entry from the RingBuffer tail (oldest) without advancing
+    /// the read cursor
+    #[inline]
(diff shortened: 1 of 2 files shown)


src/ring.rs
```diff
@@ -140,6 +140,17 @@ impl<'a> LogEntryRingBuffer<'a> {
         }
     }
 
+    /// Read a single entry from the RingBuffer tail (oldest) without advancing
+    /// the read cursor
+    #[inline]
+    pub fn peek_entry(&mut self) -> Option<LogEntry> {
```
ZackPierce, line 146: Does this need to be `&mut` access to self?

jonlamb-gh: Nope, b845234b8a7afbea06d4f377e2d08cc44db39618

PR #191, opened 2020-08-06
Add persistent epoch restart sequence counter

jonlamb-gh:


src/history.rs
```diff
@@ -164,7 +185,14 @@ impl<'a> DynamicHistory<'a> {
             clocks,
             log,
             log_items_missed: 0,
+            restart_counter: RestartSequenceCounter::new(restart_counter),
+            //_reserved_for_alignment: 0,
```
ZackPierce, line 189: Dead code

modality-probe-capi/include/probe.h
```diff
@@ -73,6 +73,13 @@ typedef struct modality_probe_causal_snapshot {
     uint16_t reserved_1;
 } modality_probe_causal_snapshot;
 
+/*
+ * Function type for retrieving the next persistent sequence number
```
ZackPierce, line 77: Could we document more here about the expectations for the generated number? Also, the circumstances in which this function will be called?

src/history.rs
```diff
@@ -164,7 +185,13 @@ impl<'a> DynamicHistory<'a> {
             clocks,
             log,
             log_items_missed: 0,
+            restart_counter: RestartSequenceCounter::new(restart_counter),
         };
+        if history.restart_counter.is_tracking_restarts() {
+            let next_persistent_epoch = history.restart_counter.next_sequence_id(history.probe_id);
```
ZackPierce, line 191: These two lines are repeated almost semantically verbatim below in `increment_local_clock`. Consider making a function which captures this shared behavior.

src/id.rs
```diff
@@ -202,6 +205,9 @@ impl EventId {
     /// The report destination buffer is too small to fit a header and/or the frontier clocks
     pub const EVENT_INSUFFICIENT_REPORT_BUFFER_SIZE: EventId =
         EventId(unsafe { NonZeroU32::new_unchecked(EventId::MAX_INTERNAL_ID - 5) });
+    /// The probe successfully initialized itself. The probe id is stored in the payload.
+    pub const EVENT_PROBE_INITIALIZED: EventId =
```
ZackPierce, line 209: Not sure why we need to store the probe id in the payload here, since recorded events are inherently localized to a probe.

src/lib.rs
```diff
@@ -237,8 +242,10 @@ impl LogicalClock {
     /// Increment the logical clock by one. If the clock portion overflows,
     /// clock wraps around and epoch is incremented. Epoch and clock both wrap
     /// around to 1.
+    ///
+    /// Returns true if the clock portion did overflow, otherwise returns false.
```
ZackPierce, line 246: To be more specific, you mean the `ticks` portion overflowing, right?

src/lib.rs
```diff
@@ -393,10 +402,15 @@ impl<'a> ModalityProbe<'a> {
...
+        restart_counter: RestartCounterProvider<'a>,
     ) -> Result<ModalityProbe<'a>, StorageSetupError> {
-        let t = ModalityProbe::<'a> {
-            history: DynamicHistory::new_at(history_memory, probe_id)?,
+        let mut t = ModalityProbe::<'a> {
+            history: DynamicHistory::new_at(history_memory, probe_id, restart_counter)?,
         };
+        t.record_event_with_payload(
+            EventId::EVENT_PROBE_INITIALIZED,
+            t.history.probe_id.get_raw(),
```
ZackPierce, line 412: As noted in `src/id.rs`, I think the payload is redundant, using log-space for no value.

src/restart_counter.rs
```diff
@@ -0,0 +1,104 @@
...
+use core::fmt;
+
+/// The persistent sequence number value returned when this
+/// probe is not tracking restarts.
+pub const NO_RESTART_TRACKING_SEQUENCE_NUMBER: u16 = 0;
+
+/// A persistent restart sequence counter
+pub trait RestartCounter {
+    /// Get the next persistent sequence number
+    fn next_sequence_id(&mut self, probe_id: ProbeId) -> u16;
```
ZackPierce, line 11: As in the C header file, I think we should talk more in the documentation about what's expected of this sequence provider. E.G. :
* Should the values returned be monotonically increasing? 
* Is there a prescribed step size? 
* How are overflows/wrap-around expected to be handled?
* Is 0 and acceptable value to return?
    * If it's not, then we may consider making this trait return `NonZeroU16` or `Option<NonZeroU16>` rather than `u16`

src/restart_counter.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl<'a> From<&'a mut dyn RestartCounter> for RestartCounterProvider<'a> {
+    fn from(r: &'a mut dyn RestartCounter) -> Self {
+        RestartCounterProvider::Rust(RustRestartCounterProvider { iface: r })
+    }
+}
+
+impl<'a> RestartCounterProvider<'a> {
+    fn next_sequence_id(&mut self, probe_id: ProbeId) -> u16 {
```
ZackPierce, line 50: Yep, the more I look at this, the more I think that `Option<NonZeroU16>` should be the return type, since it makes it clear whether we have a valid, order-able sequence number or not.  Then, the user of this function (i.e. the fn in history that deals with storing epochs during initialization or overflows) can deal with signaling further upstream this lack in whichever way is appropriate (whether that's stuffing a placeholder 0 into a ProbeEpoch or some approach)

jonlamb-gh: Makes sense to me, updated in 3e47a5b3bc3a9d66a04fd118985624f8044689c3.

PR #200, opened 2020-08-07
add write priority level field to probe

NicholasSpringer:


src/history.rs
```diff
@@ -62,6 +63,9 @@ const_assert_eq!(
 #[derive(Debug)]
 pub struct DynamicHistory<'a> {
     pub(crate) probe_id: ProbeId,
+    /// Determines the types of entries that can be written to the log at a given time
+    /// Note: currently not in use
```
ZackPierce, line 67: Should this PR be marked WIP until there's a use for this addition?

NicholasSpringer: Sure. Nathan had talked about it getting in there before the debug-collector was ready, so that potentially more work on the probe side could be done with it, but the debug collector will be coming soon anyway. I'll close the pr, since the debug collector pr can just create the field.

PR #201, opened 2020-08-10
RaceBuffer ring-buffer like API, consistent u64 cursors

NicholasSpringer:


race-buffer/model/SequenceNumbers.tla
```diff
@@ -0,0 +1,185 @@
...
+
+(* Set ProgressVal structure to be in progress *)
+SetInProgress(progress_struct) == ProgressValStruct(progress_struct.val, TRUE)
+
+-----------------------------------------------------------------------------
+
+(* Writer steps *)
+
+(* The low word of the writer is incremented until it reaches the integer maximum *)
+IncrementLow ==
```
ZackPierce, line 45: Note for reference that we should model incrementing both read and write by both 1 and 2 (due to being able to read 1 or 2 log entries).

race-buffer/src/async_reader.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        // Calculate number of entries in snapshot buffer that may have been overwritten
+        let post_overwrite_seqn = self.snap_overwrite_seqn()?;
+        let n_overwritten_in_snap = min(
+            num_missed(first_read_seqn, post_overwrite_seqn),
+            (buf_snapshot.len() as u64).into(),
+        );
+        // If any entries were missed and there is a stored prefix, then the entry after the prefix was missed.
+        // The prefix is dropped and added to the missed count (not overwritten during snap)
+        if Into::<u64>::into(n_overwritten_in_snap) > 0u64 && self.drop_prefix() {
```
ZackPierce, line 95: FWIW `u64::from(n_overwritten_in_snap)` is a more compact and equivalent way to write the conversion.

race-buffer/src/async_reader.rs
```diff
@@ -0,0 +1,286 @@
...
+        } {}
+        Ok(SeqNum::new(final_high, low))
+    }
+
+    /// Get a snapshot of the high word of a sequence number, retrying if the "updating" bit is set
+    fn snap_consistent_high<F>(snap_high: F) -> Result<u32, S::Error>
+    where
+        F: Fn() -> Result<u32, S::Error>,
+    {
+        let mut high = SeqNum::UPDATING_HIGH_MASK;
```
ZackPierce, line 174: It seems like initializing this value to something that will always return true for the first loop iteration will waste a loop iteration and check. Why not set the initial value to the result of `snap_high()?` ?

race-buffer/src/async_reader.rs
```diff
@@ -0,0 +1,286 @@
...
+        // retry at most, in the case where the first snapshot attempt and an increment to the high word happen to overlap.
+        while {
+            // Wait until the sequence number high word is not getting updated
+            initial_high = Self::snap_consistent_high(&snap_high)?;
+            low = snap_low()?;
+            final_high = snap_high()?;
+            // Check that the high word did not change between the two snapshots. If it did,
+            // The low word is unusable because we are not sure which high word it corresponds to.
+            // In that case, retry the whole sequence
+            initial_high != final_high
```
ZackPierce, line 164: I have to say, this is one of the cooler uses of "blocks are expressions" in Rust.

race-buffer/src/async_reader.rs
```diff
@@ -0,0 +1,286 @@
...
+        // If any entries were missed and there is a stored prefix, then the entry after the prefix was missed.
+        // The prefix is dropped and added to the missed count.
+        // Note: the read sequence number is updated above because the prefix was included in the last read
+        if Into::<u64>::into(n_missed_before_read) > 0 && self.drop_prefix() {
+            n_missed_before_read += 1;
+        }
+
+        // Perform reads into snapshot buffer up to write sequence number
+        let first_read_seqn = self.read_seqn;
+        let mut buf_snapshot: Vec<E> = Vec::new();
```
ZackPierce, line 80: To avoid allocating on what might be a relative hot read path and you don't want to avoid an intermediate buffer entirely in order to get bulk reads, I would recommend instead storing a `buf_snapshot` field on the `RaceReader` struct and just `clear`ing it here instead of new'ing up a new one.

race-buffer/src/buffer.rs
```diff
@@ -0,0 +1,635 @@
...
+
+impl fmt::Debug for SizeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Storage capacity must be at least {}", MIN_STORAGE_CAP,)
+    }
+}
+
+/// Round given length down to a power of 2
+#[inline]
+fn round_to_power_2(length: usize) -> usize {
```
ZackPierce, line 26: This function panics when given a value of 0. Probably ought to fix that, and add tests.

race-buffer/src/buffer.rs
```diff
@@ -0,0 +1,635 @@
...
+    /// Indicates if backing storage should be truncated to a power of 2 length
+    /// in order to use optimized indexing
+    use_base_2_indexing: bool,
+}
+
+impl<'a, E> RaceBuffer<'a, E>
+where
+    E: Entry,
+{
+    /// Create new RaceBuffer. Returns error if storage size is not power of 2
```
ZackPierce, line 56: Is this documentation about the error outdated? It seems like the function is parameterized to accept a non-power-of-2 storage-length.

race-buffer/src/buffer.rs
```diff
@@ -0,0 +1,635 @@
...
+where
+    E: Entry,
+{
+    type Item = WholeEntry<E>;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.pop()
+    }
+}
+
+/// Iteratates through a RaceBuffer
```
ZackPierce, line 294: Typo

race-buffer/src/buffer.rs
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    #[cfg(feature = "std")]
+    #[cfg(test)]
+    pub(crate) fn get_overwrite_seqn(&self) -> SeqNum {
+        self.overwrite_seqn
+    }
+}
+
+impl<E> Iterator for RaceBuffer<'_, E>
```
ZackPierce, line 284: In this case I would recommend making a dedicated iterator type that does the removal, created by a `drain(&mut self)` method on `RaceBuffer`.  This makes it extra clear that one is consuming the contents of the collection.  

For reference in the standard library: 
* https://doc.rust-lang.org/std/collections/binary_heap/struct.BinaryHeap.html#method.drain
* https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.drain

race-buffer/src/lib.rs
```diff
@@ -1,162 +1,234 @@
...
+
+impl<E> WholeEntry<E>
+where
+    E: Entry,
+{
+    /// 1 if entry is single, 2 if double
+    pub fn size(&self) -> u64 {
+        match self {
+            Self::Single(_) => 1,
+            _ => 2,
```
ZackPierce, line 173: I'd recommend against using a catch-all here in the case that the enum ever gets changed and the need to update this function gets missed.

race-buffer/src/lib.rs
```diff
@@ -1,162 +1,234 @@
...
+    /// Double entry overwritten
+    Double(E, E),
+}
+
+impl<E> WholeEntry<E>
+where
+    E: Entry,
+{
+    /// 1 if entry is single, 2 if double
+    pub fn size(&self) -> u64 {
```
ZackPierce, line 170: If the value is only 1 or 2, I'd recommend using a smaller type for the output, perhaps a u8.

race-buffer/src/lib.rs
```diff
@@ -1,162 +1,234 @@
...
-            for (idx, entry) in rbuf.iter().enumerate().filter(|(_, e)| e.is_some()) {
-                if entry.unwrap().to_index() != idx as u32 {
-                    return false;
-                }
-            }
-            return true;
-        }
+/// Calculate the number of entries missed based on the read and overwrite sequence numbers
+#[inline]
+fn num_missed(read_seqn: SeqNum, overwrite_seqn: SeqNum) -> SeqNum {
```
ZackPierce, line 139: I don't think this should return a `SeqNum`, which I understand is largely present for indexing and comparison between a few specific cursors, which is not how we would want someone to use the outcome of this function.  A `u64` seems appropriate.

ZackPierce: Nevermind, I found the places where it is being used for cursor incrementing, and that seems fine enough.

race-buffer/src/util.rs
```diff
@@ -0,0 +1,148 @@
+use super::*;
```
ZackPierce, line 1: Could we name this file something more descriptive than `util`? `test_support` maybe? `test_types` ?

race-buffer/src/buffer.rs
NicholasSpringer: I added a note and debug assertion. This function is tested in `buffer::tests::test_init_sizes`

fenced-ring-buffer/src/async_reader.rs
```diff
@@ -71,47 +74,47 @@ where
...
         // The prefix is dropped and added to the missed count.
         // Note: the read sequence number is updated above because the prefix was included in the last read
-        if Into::<u64>::into(n_missed_before_read) > 0 && self.drop_prefix() {
+        if u64::from(n_missed_before_read) > 0 && self.drop_prefix() {
             n_missed_before_read += 1;
         }
 
         // Perform reads into snapshot buffer up to write sequence number
         let first_read_seqn = self.read_seqn;
-        let mut buf_snapshot: Vec<E> = Vec::new();
```
ZackPierce: It is probably wise / necessary to call `self.buf_snapshot.clear()` here because the `clear` call near the end of the function may not be called if there is an early-return (e.g. if an error is returned via a `?`).

src/log.rs
```diff
@@ -1,13 +1,13 @@
...
-use race_buffer::RaceBuffer;
+use fenced_ring_buffer::FencedRingBuffer;
 
 pub(crate) const CLOCK_MASK: u32 = 0b1000_0000_0000_0000_0000_0000_0000_0000;
 pub(crate) const EVENT_WITH_PAYLOAD_MASK: u32 = 0b0100_0000_0000_0000_0000_0000_0000_0000;
 
-/// RaceBuffer used to store log entries at each probe
-pub type RaceLog<'a> = RaceBuffer<'a, LogEntry>;
+/// FencedRingBuffer used to store log entries at each probe
+pub type RaceLog<'a> = FencedRingBuffer<'a, LogEntry>;
```
ZackPierce, line 10: For consistency, should change this type name too.

PR #203, opened 2020-08-14
Use new reporting format for graph building and build them on the fly

### LINKED ISSUE #196: Reimplement graph builders to be used "live" / incrementally
The graph builders need to be reworked for the new report format. While we're here, we should also see if we can move most if not all of the fold state into the `GraphBuilder` traits. This would allow the builders to be used live, as data is coming in _or_ on a complete data set. If this is possible, then the graph builders become useful as a library.

### LINKED ISSUE #204: Use row-based JSON rather than CSV for collected reports
The CSV representation of the log is a little cumbersome due to the emptiness or presence of some fields denoting the row's type. If we go row-by-row with JSON objects, the rows can be deserialized directly into their type and do not have to adhere to a file-wide fixed schema.

pittma:
Closes #196.
Closes #204.


collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -134,20 +141,134 @@ impl From<EventLogEntry> for LogEntryData {
...
-    tc_probe_clock: Option<u16>,
+pub struct LogFileRow {
+    pub session_id: u32,
+    pub sequence_number: u64,
+    pub sequence_index: u32,
+    pub receive_time: DateTime<Utc>,
+    pub probe_id: u32,
+    pub fc_probe_id: Option<u32>,
+    pub fc_probe_epoch: Option<u16>,
+    pub fc_probe_clock: Option<u16>,
```
ZackPierce, line 152: Is there a reason we have ungrouped Option fields as opposed to semantically-linked groups? E.G.`pub fc: Option<FrontierClock>,` rather than 3 flat fields (now that we're not in csv-land?).

Actually, if we're okay with nested structure now, how about using an actual enum for switching between the 3 main content variants?

`enum RowContents { FrontierClock(etc), Event(etc), TraceClock(etc) }`

pittma:
Nope! This type is actually not used anymore. It's going to come out
altogether as soon as I have time to fix my graph tests that are still
using it. I've a unchecked item on the PR to remind myself to do this.

collectors/modality-probe-collector-common/src/lib.rs
pittma: Nope! This type is actually not used anymore. It's going to come out altogether as soon as I have time to fix my graph tests that are still using it. I've a unchecked item on the PR to remind myself to do this. Everything is `ReportLogEntry` now, which does exactly that!

ZackPierce: Excellent, good to know.

collectors/modality-probe-collector-common/src/json.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn from(e: &serde_json::error::Error) -> Error {
+        Error::Serialization(format!("json failure: {}", e))
+    }
+}
+
+pub fn read_log_entries<R: Read>(r: &mut R) -> Result<Vec<ReportLogEntry>, Error> {
+    let br = BufReader::new(r);
+    let entries: Result<Vec<ReportLogEntry>, _> = br
+        .lines()
+        .into_iter()
```
ZackPierce, line 31: It seems unlikely this `into_iter()` is necessary.

collectors/modality-probe-collector-common/src/lib.rs
ZackPierce: Recommend marking this struct as `#[cfg(test)]` in the meantime.

collectors/modality-probe-udp-collector/src/opts.rs
```diff
@@ -44,7 +44,7 @@ impl From<Opts> for Config {
             output_file: o.output_file.unwrap_or_else(|| {
                 std::env::current_dir()
                     .expect("Could not retrieve current directory")
-                    .join(format!("session_{}_log_entries.csv", session_id))
+                    .join(format!("session_{}_log_entries.mtf", session_id))
```
ZackPierce, line 47: Multi-Track-File? If we're doing line-separated JSON, then the recommended practice is naming it `.jsonl`, I think. (Per: http://jsonlines.org/ )

modality-probe-cli/src/export/graph.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+#[derive(Default)]
+pub(super) struct NodeAndEdgeLists<G, W> {
+    nodes: HashMap<G, W>,
+    edges: HashMap<(G, G), W>,
+}
+
+impl NodeAndEdgeLists<GraphEvent, ()> {
+    pub fn into_interactions(self) -> NodeAndEdgeLists<(ProbeId, LogicalClock), u32> {
```
ZackPierce, line 65: It seems like this and the following 2 functions are read-only for self, and could easily take `&self` rather than consuming `self`. Any particular reason we're consuming here?

modality-probe-cli/src/export/graph.rs
```diff
@@ -0,0 +1,512 @@
...
+        for ((s, t), _) in self.edges.into_iter() {
+            let weight = edges
+                .entry(((s.probe_id, s.clock), (t.probe_id, t.clock)))
+                .or_insert(0);
+            *weight += 1;
+        }
+        NodeAndEdgeLists { nodes, edges }
+    }
+
+    pub fn into_states(self) -> NodeAndEdgeLists<(ProbeId, EventId), u32> {
```
ZackPierce, line 82: Could we have a bit of documentation here on what the states mean here?

ZackPierce: At a glance, it seems like this is making something like a non-normalized Markov chain representation of event transitions.

modality-probe-cli/src/export/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+        (GraphType::Cyclic, false) => println!("{}", graph.graph.into_states().to_dot(&cfg)?),
+        (GraphType::Cyclic, true) => println!("{}", graph.graph.into_topology().to_dot(&cfg)?),
+    }
+
+    Ok(())
+}
+
+struct Cfg {
+    pub probes: HashMap<u32, ProbeMeta>,
+    pub events: HashMap<(Uuid, u32), EventMeta>,
```
ZackPierce, line 109: It seems like this representation might fail to capture events that arrived from a probe not associated with a pre-known component.

modality-probe-cli/src/export/graph.rs
```diff
@@ -0,0 +1,512 @@
...
+    fn complete_dot() {
+        let cfg = cfg();
+        let diamond_log = modality_probe_graph::test_support::diamond()
+            .into_iter()
+            .map(|e| (&e).try_into().unwrap())
+            .peekable();
+        let graph = super::log_to_graph(diamond_log).unwrap();
+
+        let dot = graph.graph.to_dot(&cfg).unwrap();
+        assert!(dot.contains("one_1_0 -> two_1_2"), dot);
```
ZackPierce, line 471: Is there an approach we could use for validating the output dot content's syntax/structure?

modality-probe-cli/src/export/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+
+use modality_probe_collector_common::{json, Error as CollectorError};
+
+use crate::{component::Component, events::Events};
+
+mod graph;
+use graph::{EventMeta, ProbeMeta};
+
+/// Export a textual representation of a causal graph using the
+/// collected coumnar form as input.
```
ZackPierce, line 18: Typo: columnar was probably intended

modality-probe-cli/src/export/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// This can be either `cyclic` or `acyclic`.
+    ///
+    /// * A cyclic graph is one which shows the possible paths from
+    /// either an event or a probe.
+    ///
+    /// * An acyclic graph shows the causal history of either all
+    /// events or the interactions between traces in the system.
+    #[structopt(required = true)]
+    pub graph_type: GraphType,
```
ZackPierce, line 42: I don't know if CLI-API is up for review/discussion, but I am somewhat surprised by the cyclic/acyclic terminology here.  In practice, this flag seems to control the difference between producing a causal history or a summarized/collapsed version of the state-transitions that occurred in such a history.

nathanaschbacher:
It's up for changing.

modality-probe-cli/src/export/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+    })?;
+
+    let graph = graph::log_to_graph(
+        json::read_log_entries(&mut log_file)?
+            .into_iter()
+            .peekable(),
+    )?;
+
+    let mut out = String::new();
+    writeln!(out, "Digraph G {{")?;
```
ZackPierce, line 95: I think this `writeln!` might be unnecessary, given that we now write an identical line (save capitalization) in each of the `to_dot` like methods invoked below.

modality-probe-cli/src/export/mod.rs
```diff
@@ -0,0 +1,340 @@
...
+            line: ie.line,
+        };
+        events.insert((nil_uuid, ie.id.0), ev.clone());
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod test {
```
ZackPierce, line 196: A high-level test here (or an external integration test) demonstrating that each combination of the CLI options can produce valid, non-empty graphviz-compatible output would be most awesome.

collectors/modality-probe-udp-collector/src/opts.rs
pittma: This is a @nathanaschbacher-ism. It's "modailty trace file". I'm happy to change it to whatever, though.

modality-probe-graph/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "modality-probe-graph"
+version = "0.1.0"
+authors = ["dan pittman <dan@auxon.io>"]
+edition = "2018"
+
+[features]
+test_support = ["chrono"]
```
ZackPierce, line 8: I recommend making the default features explicit by including an entry for it with an empty array, i.e. `default = []`

modality-probe-cli/src/export/graph.rs
pittma: I guess you could say it was vestigial, an earlier pass on this the output `G` type parameter was still `GraphEvent`. This isn't intended to be used outside here, so I just left it. It was fine to consume them.

modality-probe-cli/src/export/mod.rs
pittma: That is true. In a very early pass (pre-report redux), I just used "`UNKNOWN_EVENT`" in the metadata, I can go back that direction here.

modality-probe-cli/src/export/graph.rs
pittma: Yeah, the thing you found below about the double `digraph` means this is obviously broken. I have not gotten to spend as much time testing here as I would like. Time willing I'd find a dot library and do some round-trips, I think.

modality-probe-cli/src/export/mod.rs
pittma: yeesh, and it's also not true anymore!

modality-probe-cli/src/export/mod.rs
pittma: Yeah, those were @nathanaschbacher's too. Again, I don't have a strong opinion on what to call these. The "cyclic" ones are event or probe oriented while the "acyclic" ones are based on causality.

modality-probe-graph/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+/// Errors returned by the `EventDigraph` methods.
+#[derive(Debug, Error)]
+pub enum Error {
+    /// Encountered an error when using the writer when exporting a
+    /// graph in a textual format.
+    #[error(display = "Formatting graph failed: {}", _0)]
+    Io(String),
+    /// The graph builders iterate over a `Result` to leave room for
+    /// deserialization; if a builder encounters a `Err`, this error
+    /// is returned with that error's `display` implementaion inside.
```
ZackPierce, line 41: Typo: implementation

modality-probe-cli/src/export/mod.rs
pittma: Totally agree. There is not enough testing here at all.

modality-probe-graph/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+/// A type that eats Modality Probe reports and stuffs them into its
+/// inner graph type.
+#[derive(Debug, PartialEq, Eq)]
+pub struct EventDigraph<G: Graph> {
+    /// The EventDigraph retains ownership of the inner graph, but
+    /// must make it available so that its purpose-built functionality
+    /// is accessible.
+    pub graph: G,
+
+    // When a foreign trace clock is the last thing in a report chunk,
```
ZackPierce, line 58: I think using `///` is the right choice for fields, private or public. This allows switching field visibility without accidentally omitting the docstrings because the preceding lines didn't get an extra-slash added.

modality-probe-graph/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+                    }
+                    for lc in pending_edges.iter() {
+                        if let Some(e) = self.last_event_by_probe_and_clock.get(lc) {
+                            self.graph.add_edge(*e, node);
+                        }
+                    }
+                    pending_edges.clear();
+                    prev_event = Some(node);
+                }
+                EventLogEntry::EventWithPayload(id, payload) => {
```
ZackPierce, line 136: Looks like the handling between Event and EventWithPayload is almost entirely duplicative, with the exception of intialization of the GraphEvent payload field. Non-essential, but it would be nice to collapse them down.

modality-probe-graph/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+                            self.graph.add_edge(*e, node);
+                        }
+                    }
+                    pending_edges.clear();
+                    prev_event = Some(node);
+                }
+                EventLogEntry::TraceClock(lc) => {
+                    if lc.id == probe_id {
+                        // when we see a clock, the previous event we
+                        // saw is the last event from the previous
```
ZackPierce, line 174: Is this true even across Report boundaries when the Reports arrive out of order?

modality-probe-cli/src/export/mod.rs
ZackPierce: An alternate approach is to use an enum for keying instead of a Uuid. Maybe like:

```rust
enum {
    Component(Uuid),
    Internal,
    UnknownAtProbe(ProbeId),
}
```

modality-probe-cli/src/export/mod.rs
pittma: Marking this one as resolved for now, as we're not sure what to permanently call this yet.

modality-probe-graph/src/lib.rs
pittma: It's not true across report boundaries, but this is all within a report boundary. This function just takes a single report "chunk" at a time.

ZackPierce: I'm concerned about the state consistency across reports, since this function deals with cross-report state stored in the `EventDigraph`. Here, I'm asking whether ` self.last_event_by_probe_and_clock` collection's consistency and assumed invariants hold when `EventDigraph::add_report` is called with reports that have arrived out of order.

pittma: >  Here, I'm asking whether self.last_event_by_probe_and_clock collection's consistency and assumed invariants hold when EventDigraph::add_report is called with reports that have arrived out of order.

IIUYC, yes, they do. By

> but this is all within a report boundary

I meant that `prev_event` is only valid for a single call of `add_report`. It's initialized just before the loop, so this insert will only happen if there's an event from the previous iteration, within the current invocation of `add_report`. Therefore when we see a trace clock, but we've seen an event before **in the same report**, then it's safe to assume that event is the last event from the previous clock span. You're right in thinking that, across a report boundary, this could be inconsistent—in particular if the first entry in a report is a trace clock and `prev_event` had a dangling value. It's that `prev_event` is used within a single report that makes this work.

pittma: Thinking this through a bit further, I think this does leave a hole when we can do better. If the first entry in a report is a self trace clock, we'll not try and lookup the last event from the previous sequence to add it to our `last_event_by_probe_and_clock`. Updating this now.

PR #209, opened 2020-08-21
Debug collector

NicholasSpringer:


collectors/modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,550 @@
...
+    #[structopt(short = "s", long = "session-id", default_value = "0")]
+    session_id: u32,
+
+    /// Specifies 32 bit architecture of target system
+    #[structopt(short = "32", long = "32-bit", conflicts_with = "word-size-64")]
+    word_size_32: bool,
+
+    /// Specifies 32 bit architecture of target system
+    #[structopt(short = "64", long = "64-bit")]
+    word_size_64: bool,
```
ZackPierce, line 43: Can't we parse this out of the elf file header?

ZackPierce: I suppose if we allow the system architecture to differ from the elf file, this serves a value.

collectors/modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// Raw addresses should be in hex format, prefixed with '0x' or '0X'
+    /// Probe pointer addresses and symbols should be prefixed with `*`.
+    /// Ex: *0X100 0x104 symbol1 *symbol2 0X108 *0x10c
+    probe_syms: Vec<String>,
+}
+
+/// Turn CLI options into configuration for the collector
+pub(crate) fn config_from_options(options: CLIOptions) -> Result<Config, CLIError> {
+    let mut probe_addrs = Vec::new();
+    let mut symbols = Vec::new();
```
ZackPierce, line 80: Minor style nit - these two variables should probably be declared immediately prior to their point of first use.  Shortening the distance  between declaration, initialization, and use reduces the surface area for accidental misuse and reduces no-value scrollback for context discovery.

collectors/modality-probe-debug-collector/src/cli.rs
```diff
@@ -0,0 +1,550 @@
...
+    use std::format;
+    use std::fs::canonicalize;
+    use std::process::Command;
+    use std::str::from_utf8;
+    use std::str::FromStr;
+    use std::time::Duration;
+
+    const SYMBOLS_32_BIN_PATH: &str =
+        "./tests/symbols-example/target/thumbv7em-none-eabihf/debug/symbols-example";
+    const EMPTY_BIN_PATH: &str = "./tests/empty-example/target/debug/empty-example";
```
ZackPierce, line 222: As CI noted, this probably ought to be cfg-gated to the targets where it is in use.

collectors/modality-probe-debug-collector/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+fn main() {
+    let opts = CLIOptions::from_args();
+    let config = match config_from_options(opts) {
+        Ok(cfg) => cfg,
+        Err(err) => {
+            println!("{}", err);
+            return;
+        }
+    };
+    let (_, shutdown_receiver) = channel();
```
ZackPierce, line 18: It looks like the piping is incomplete here, since there's nothing holding on to the shutdown channel's sender.  I presume this is supposed to be hooked up to a control-C type hook.

src/history.rs
```diff
@@ -48,56 +49,65 @@ const_assert_eq!(4, align_of::<ModalityProbeInstant>());
...
+        + size_of::<u64>(),
     size_of::<DynamicHistory>()
 );
 
 /// Manages the core of a probe in-memory implementation
 /// backed by runtime-sized arrays of current logical clocks
 /// and probe log items
+///
+/// Note: overwrite_priority, probe_id, and log must be accessed by the debug collector using direct memory access,
+/// so they must be public fields (in order for the debug collector to calculate their offsets).
```
ZackPierce, line 83: Rather than making these fields public, could we instead expose the offsets through some feature-flag-gated functions in modality-probe?  Seems like that would be less prone to the sundry issues that come with exposing struct internals completely.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+        .get_byte_offset() as u64
+}
+
+fn log_storage_addr_offset() -> u64 {
+    offset_of!(DynamicHistory => log: FencedRingBuffer<LogEntry> => storage).get_byte_offset()
+        as u64
+}
+
+fn log_storage_cap_offset(n_word_bytes: u8) -> u64 {
+    // Assume storage addr is a u32
```
ZackPierce, line 113: What's the basis for this assumption?  Is it one we can/should check at runtime?

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+    LogProcessingError,
+    #[error(display = "Error serializing the report")]
+    ReportSerializationError,
+    #[error(display = "Error using output file")]
+    FileError,
+}
+
+/// Trait used to specify backend used to access device memory
+pub trait MemoryAccessor {
+    fn read_word(&mut self, addr: Word) -> Result<Word, DebugCollectorError>;
```
ZackPierce, line 159: It seems like the implementations of this trait only ever return the `DebugCollectorError::TargetError` variant of `DebugCollectorError`. I recommend changing the error-branch return type of this trait to something specific to the kinds of failures that are actually expected to occur in the trait implementations. Then, as necessary, at the call site or in a `From` implementation, map that trait-specific error return to the appropriate higher-level error type (i.e. `DebugCollectorError`) variants.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+            let probes = probe_rs::Probe::list_all();
+            if probes.is_empty() {
+                return Err(DebugCollectorError::TargetError);
+            }
+            let probe = probes[0]
+                .open()
+                .map_err(|_e| DebugCollectorError::TargetError)?;
+            let session = probe
+                .attach(target)
+                .map_err(|_e| DebugCollectorError::TargetError)?;
```
ZackPierce, line 417: The lack of detailed information in this set of `TargetError` instances (compared to the other ways a `DebugCollectorError::TargetError` might get tossed in this system) would make remediation of the error tricky, I imagine. Is there some further DebugCollectorError variants or sub-errors nested within a DebugCollectorError variant that could be developed to make a correction path apparent to the end user?

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+        let buf_addr = mem_borrow.read_word(hist_addr + log_storage_addr_offset())?;
+        let buf_cap = mem_borrow
+            .read_word(hist_addr + log_storage_cap_offset(hist_addr.size()))?
+            .into();
+        let write_seqn_high_addr = hist_addr + write_seqn_high_offset();
+        let write_seqn_low_addr = hist_addr + write_seqn_low_offset();
+        let overwrite_seqn_high_addr = hist_addr + overwrite_seqn_high_offset();
+        let overwrite_seqn_low_addr = hist_addr + overwrite_seqn_low_offset();
+        let priority_field_addr = hist_addr + overwrite_priority_offset();
+        drop(mem_borrow);
```
ZackPierce, line 310: Rather than using `drop` I recommend instead using a block to nest the borrow and the subsequent actions with it, returning the retrieved values from the block.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1169 @@
...
+
+    fn merge_clock(clocks: &mut Vec<LogicalClock>, ext_clock: LogicalClock) {
+        let mut existed = false;
+        for c in clocks.iter_mut() {
+            if c.id == ext_clock.id {
+                if OrdClock(ext_clock.epoch, ext_clock.ticks) > OrdClock(c.epoch, c.ticks) {
+                    c.epoch = ext_clock.epoch;
+                    c.ticks = ext_clock.ticks;
+                }
+                existed = true;
```
ZackPierce, line 390: I think that we're likely to be maintaining only a single instance of a logical clock per probe id, so it should be fine to `break` after this point.

fenced-ring-buffer/src/buffer.rs
```diff
@@ -0,0 +1,660 @@
+//! The writer of the FencedRingBuffer, which contains the actual buffer. Entries can be synchronously written and read/iterated
+//! from this struct.
+use crate::{get_seqn_index, num_missed, Entry, SeqNum, WholeEntry};
+//use core::iter::Iterator;
```
ZackPierce, line 4: Dead code, should be removed.

src/history.rs
```diff
@@ -48,56 +49,65 @@ const_assert_eq!(4, align_of::<ModalityProbeInstant>());
...
-    pub(crate) report_seq_num: u64,
     pub(crate) restart_counter: RestartSequenceCounter<'a>,
+    pub(crate) report_seq_num: u64,
 }
 
 #[derive(Debug)]
 struct ClocksFullError;
 
 impl<'a> DynamicHistory<'a> {
+    /// Create new DynamicHistory at given destination
```
ZackPierce, line 110: Technically, it'll be put somewhere inside the destination, but it's not certain that the `DynamicHistory` structure will start exactly at the beginning of the provided destination byte array, thanks to alignment-checking done in `fixed_slice_ve::single::embed`.

src/lib.rs
```diff
@@ -332,7 +331,8 @@ pub trait Probe {
 /// * Can produce and merge transparent snapshots
 #[repr(C)]
 pub struct ModalityProbe<'a> {
-    history: &'a mut DynamicHistory<'a>,
+    /// Publicly accessible for direct access by debug-collector
+    pub history: &'a mut DynamicHistory<'a>,
```
ZackPierce, line 335: As noted elsewhere, it seems like we ought to prefer not making the probe internals public (and publicly mutable) if the main goal for doing so is the determination of field offsets for the debug collector (and its opt-in risk profile).

src/history.rs
```diff
@@ -762,10 +744,10 @@ mod test {
 
         // One more to get the remainder
         let bytes_written = h.report(&mut report_dest).unwrap().unwrap();
-        assert_eq!(bytes_written.get(), 59);
+        assert_eq!(bytes_written.get(), 51);
         let log_report = WireReport::new(&report_dest[..bytes_written.get()]).unwrap();
         assert_eq!(log_report.n_clocks() as usize, h.clocks.len());
-        assert_eq!(log_report.n_log_entries(), 6);
+        assert_eq!(log_report.n_log_entries(), 4);
```
ZackPierce, line 750: Why did this change?

collectors/modality-probe-debug-collector/src/cli.rs
NicholasSpringer: Also, as it currently stands, the user does not necessarily need to supply an elf file.

src/history.rs
NicholasSpringer: The history struct is larger than it used to be, causing the log slice to be decreased by 2 entries in this test, meaning there are 2 less entries written at the start of the test.

collectors/modality-probe-debug-collector/src/lib.rs
NicholasSpringer: Sorry, this comment was outdated, `n_word_bytes` is 4 or 8 depending on the target word size.

NicholasSpringer:
Still need to add run-time sanity checking to ensure there is actually a probe where we think

Cargo.toml
```diff
@@ -14,23 +14,27 @@ readme = "README.md"
 
 [features]
 default = []
-std = ["serde", "proptest", "schemars", "proptest-derive"]
+std = ["serde", "proptest", "schemars", "proptest-derive", "field-offset"]
```
ZackPierce, line 17: Could we hide the field-offset dependency behind a distinct feature flag, please?  IIRC, that crate is no-std friendly and the capabilities related to having it are distinct from the use cases enabled by the other things currently behind the "std" flag.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -0,0 +1,1345 @@
...
+            }
+        }
+    }
+
+    fn scan_guard_bytes(
+        mut addr: Word,
+        mem_accessor: Rc<RefCell<dyn MemoryAccessor>>,
+    ) -> Result<Word, Error> {
+        let mut mem = mem_accessor.borrow_mut();
+        loop {
```
ZackPierce, line 372: This loop ought to be bounded by a constant.  One candidate off the top of my head is the alignment-size of the ModalityProbe (or a small multiple thereof).

src/lib.rs
```diff
@@ -316,10 +325,21 @@ impl<'a> ModalityProbe<'a> {
         probe_id: ProbeId,
         restart_counter: RestartCounterProvider<'a>,
     ) -> Result<&'a mut ModalityProbe<'a>, StorageSetupError> {
+        // Align memory before embedding so that all padding is filled with guard bytes
+        let padding_offset = memory.as_ptr().align_offset(align_of::<Self>());
```
ZackPierce, line 329: Not important for this PR, but this sort of guarding seems like a pattern that could be moved in a general way into a variant of the `embed` function (currently provided by the fixed-slice-vec crate), e.g:

```
pub fn embed_with_guard_prefix<'a, T, F, E>(destination: &'a mut [u8], guard_byte: u8, f: F) -> Result<&'a mut T, EmbedValueError<E>>
where
    F: FnOnce(&'a mut [u8]) -> Result<T, E>
```

src/lib.rs
```diff
@@ -267,10 +268,18 @@ pub trait Probe {
 /// * Can produce and merge transparent snapshots
 #[repr(C)]
 pub struct ModalityProbe<'a> {
+    fingerprint: u32,
+    /// Pad fingerprint up to 8 bytes for consistent history offset
```
ZackPierce, line 272: Here "consistent" means "with the same field offset on both 32-bit and 64-bit architectures", right?

ZackPierce: If so, recommend adding some detail to that effect in the doc-comment.

src/lib.rs
NicholasSpringer: I'll add a note as a future improvement

PR #225, opened 2020-09-08
Debug collector gdb support

NicholasSpringer:
Add ability to connect to a GDB server to read target memory instead of directly connecting to the target chip using probe-rs. Has history of #209.


jonlamb-gh:
Will migrate this over to the commercial repo if needed

PR #229, opened 2020-09-10
Add ::MIN and impl Default for ProbeEpoch and ProbeTicks types

mullr:


jonlamb-gh (approved):

nathanaschbacher:
This still needs more discussion according to @mullr

nathanaschbacher (changes requested):
@mullr tells me that this actually needs more discussion, so I'm marking that with a request for changes... namely more discussion.

pittma:
This is subsumed in #242 by https://github.com/auxoncorp/modality-probe/pull/242/commits/c29560907307eba1411370b92c296225896ed77f.

PR #241, opened 2020-09-15
Fix C API reporting return code

jonlamb-gh:


ZackPierce (approved):
Good work, and thanks for adding a regression check.

PR #248, opened 2020-09-18
Do not allow self-merges, reorganize epoch calculation

ZackPierce:
* When merging, ignore snapshots with the same probe id as the merging probe.
* Only insert events relating to probe epoch calculation *after* the initial clock set gets written to the log
  * Make sure the PROBE_INITIALIZED event is the first event after the clocks in the log for a new probe
* Add some tests to confirm when and how we expect clock-ticks to be reflected in snapshots and now-retrieved instants.
* Put the next-epoch calculation for the probe in one spot,  `calculate_next_epoch` , with more limited mutability.
* Dispel magic on the 0-value for probe epochs as seen through the lens of the persistent restart counter interfaces
* Add an explicit error path for the restart counter interfaces

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -55,7 +55,7 @@ const_assert_eq!(
         + size_of::<u32>()
         + size_of::<LogicalClock>()
         + size_of::<FixedSliceVec<'_, LogicalClock>>()
-        + size_of::<RestartSequenceCounter<'_>>()
+        + size_of::<RestartCounterProvider<'_>>()
         + size_of::<u64>(),
     size_of::<DynamicHistory>()
 );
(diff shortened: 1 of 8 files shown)


src/history.rs
```diff
@@ -219,17 +225,35 @@ impl<'a> DynamicHistory<'a> {
...
+                )
+            } else {
+                (
+                    ProbeEpoch::MIN,
+                    RestartCounterProvidedInvalidEpochSeqId(true),
+                )
             }
+        } else if let Some(prior_epoch) = prior_epoch {
+            (
+                ProbeEpoch(core::cmp::max(1, prior_epoch.0.wrapping_add(1))),
```
mullr, line 249: Shouldn't this one go away too?

mullr: that is, I think this should just be regular wrapping add, now that the `1` epoch is no longer special.

ZackPierce: Addressed in the latest commit

mullr (approved):
lgtm after the build is fixed

PR #255, opened 2020-09-28
cli: exclude internal events in graph output by default

### LINKED ISSUE #235: During export, exclude internal reserved events by default, and add a flag to re-include them
When the Modality reserved internal events are included in the DOT export, they can make the trace a bit harder to understand from the user's perspective. Especially on the cyclic-style export where the events are like states in a state graph. Because pretty much everything eventually "publishes a report" it has the effect in the graph of making it look like basically everything is connected to the publish report event. 

I think this would be helped a little bit when these events (despite sharing the same reserved ID) are separated and grouped in the cluster for the Probe they happened in, but still these should be only optionally included in the export.

My suggestion is to suppress Modality reserved events from the export by default, and add a flag like `--with-internal-events` or `--include-internal-events` or something which then has the export behave more or less as it currently does.

pittma:
This PR adds the `--include-internal-events` switch to the `export` sub-command. By default, these are now excluded.

---

Closes #235.

### DIFF
--- a/modality-probe-cli/src/export/graph.rs
+++ b/modality-probe-cli/src/export/graph.rs
@@ -69,8 +69,11 @@ where
 }
 
 impl NodeAndEdgeLists<GraphEvent> {
-    pub fn as_complete<'a>(&'a self) -> NodeAndEdgeLists<&'a GraphEvent> {
-        self.filter(|_| true, |_, _| true)
+    pub fn as_complete<'a>(&'a self, include_internals: bool) -> NodeAndEdgeLists<&'a GraphEvent> {
+        self.filter(
+            |n| include_internals || !n.id.is_internal(),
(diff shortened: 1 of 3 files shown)


jonlamb-gh (approved):

jonlamb-gh:
I noticed an issue today running `export` over this [log.jsonl](https://github.com/auxoncorp/modality-probe/files/5301234/trace_log.txt). It looks like we're missing some arrows when `--include-internal-events` is omitted.

With internal events:
cyclic_with_internals

Without internal events:
cyclic_without_internals

jonlamb-gh (changes requested):
Can you look into the missing arrows issue?

pittma:
@jonlamb-gh I'm having a hard time seeing what's missing…

pittma:
screen_2020-09-30-12-48-23

jonlamb-gh (approved):

PR #258, opened 2020-09-30
Dependencies cleanup

ZackPierce:
* Removes the `proptest_derive` dependency which was causing us to pull in a whole duplicate of the `syn`/`quote` proc-macro ecosystem
* Bump the proptest dependency, removing a duplicate of the `rand` ecosystem of dependencies
* Fixes our `Arbitrary` impl for ProbeId
* Bump the probe-rs dep to remove our last duplicate transitive dependency (which was on rusb)

This cleanup has been brought to you by the command `cargo tree --workspace -d`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 
 [features]
 default = []
-std = ["serde", "proptest", "schemars", "proptest-derive"]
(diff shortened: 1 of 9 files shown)


jonlamb-gh (approved):
:+1:

PR #261, opened 2020-10-01
Add description formatting utilities

### LINKED ISSUE #259: Add support for a format specifier in event description strings to format a payload value when displayed

jonlamb-gh:
Adds a trait providing description string payload formatting.

Closes #259


nathanaschbacher:
What does this look like to use from both Rust and C? Does this also include the "glue" to the code to do this when the manifest is consumed to decorate the raw data with more context?

jonlamb-gh:
> What does this look like to use from both Rust and C? Does this also include the "glue" to the code to do this when the manifest is consumed to decorate the raw data with more context?

Check out the diff in this PR, I added a formatted description to both the C and the Rust examples. This only adds the capability to format a dynamic string with a payload value, was thinking it'd be glue'd together in #262 as part of `log`.

pittma:
> > What does this look like to use from both Rust and C? Does this also include the "glue" to the code to do this when the manifest is consumed to decorate the raw data with more context?
> 
> Check out the diff in this PR, I added a formatted description to both the C and the Rust examples. This only adds the capability to format a dynamic string with a payload value, was thinking it'd be glue'd together in #262 as part of `log`.

This is what I'm thinking as well, and it's why I left the `log` PR as a draft. The only additional thought I had about this specifically is that I wonder if it would be useful to really lean into the logging disposition and add log level-specific definitions that map to severity levels. Something like you get with Rust logging with `log::info!(…)`.

nathanaschbacher:
> > > What does this look like to use from both Rust and C? Does this also include the "glue" to the code to do this when the manifest is consumed to decorate the raw data with more context?
> > 
> > 
> > Check out the diff in this PR, I added a formatted description to both the C and the Rust examples. This only adds the capability to format a dynamic string with a payload value, was thinking it'd be glue'd together in #262 as part of `log`.
> 
> This is what I'm thinking as well, and it's why I left the `log` PR as a draft. The only additional thought I had about this specifically is that I wonder if it would be useful to really lean into the logging disposition and add log level-specific definitions that map to severity levels. Something like you get with Rust logging with `log::info!(…)`.

I like that last bit, and I could totally see that sort of functionality as part of the "Log friendly API effort" here #264

nathanaschbacher:
I think what needs to happen to merge this in as a first pass is to update the `export` capability to glue together the SUT event description w/ the payload in the output.

pittma:
> I think what needs to happen to merge this in as a first pass is to update the `export` capability to glue together the SUT event description w/ the payload in the output.

I can take a stab at that once the `log` design settles.

pittma:
> I can take a stab at that once the `log` design settles.

https://github.com/auxoncorp/modality-probe/pull/261/commits/f8db997a05fa78d57ca802e8a83ffff91292ed6d interpolates payloads into descriptions if the description contains `{}`:

```
        PRODUCER_MEASUREMENT_SAMPLED_PRODUCER_PROBE_0_3 [
                label        = "PRODUCER_MEASUREMENT_SAMPLED"
                
                message      = "Measurement producer sampled a value for transmission: 2"
                
                file         = "rust-example/src/main.rs" 
                probe        = "PRODUCER_PROBE"
                tags         = "producer;measurement sample"
                
                payload      = 2                      
                                                 
                raw_event_id = 2
                raw_probe_id = 837318897  
        ];
```

modality-probe-cli/src/visualize/graph.rs
```diff
@@ -219,13 +219,26 @@ fn graph_to_tree<'a>(
...
-                meta::parsed_payload(emeta.type_hint.as_ref().map(|s| s.as_ref()), node.payload)
-                    .unwrap_or(None);
+            let payload = node.payload.and_then(|pl| {
+                emeta.type_hint.as_ref().and_then(|th| {
+                    meta::parsed_payload(Some(th.as_ref()), Some(pl))
+                        .ok()
+                        .flatten()
+                })
+            });
+            let has_log_str = emeta.description.contains("{}");
```
jonlamb-gh, line 229: The `DescriptionFormat` trait has a `contains_formatting()` method for this.
Should make it easier for us to expand later on for extra formatting, like `{:X}` for hex.

modality-probe-cli/src/visualize/graph.rs
```diff
@@ -219,13 +219,26 @@ fn graph_to_tree<'a>(
...
+            let log_str = if has_log_str && payload.is_some() {
+                Some(emeta.description.replace("{}", payload.as_ref().unwrap()))
```
jonlamb-gh, line 231: Maybe this should be using the `format_payload()` method from the `DescriptionFormat` trait?

pittma: Yep! just made this change. Pushing shortly.

PR #262, opened 2020-10-02
log sub-command

### LINKED ISSUE #260: a `log` sub-command
Add a sub-command to the modality probe CLI that turns a report into something `git log` like, with a few options:

* A single probe's timeline.
* A complete timeline, interleaved by probe and/or some other interleaving options.

pittma:
Closes #260.


pittma:
@jonlamb-gh I think what I'd like to do is follow up with the inclusion of the log formatting _after_ the dust settles with this and @nathanaschbacher's desires on what this command looks like.

PR #263, opened 2020-10-04
fix Rust toolchain link

jeffweiss:
Prior to this commit the README referenced `rustup.sh` as the canonical
host for installing Rust; however, the proper host is instead
`rustup.rs`, with the subsequent script using `sh.rustup.rs`. The commit
fixes the link for users following the modality-probe steps on a clean
system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ environments.
 
 ### Dependencies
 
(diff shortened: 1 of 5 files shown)


jonlamb-gh (approved):
Thanks!

PR #285, opened 2020-10-26
Add common time domain support in the probe

jonlamb-gh:
This commit incorporates a common wall clock time domain into the probe.

* Wall clock time is represented in terms of 61-bit nanoseconds.
* Constructing a probe now takes two additional arguments: a time resolution in ns, and a wall clock id
* Adds several new APIs that add `with_time` variants of event recording as well as a standalone `record_time()` method
* Adds a new `LogEntry` variant for wall clock time: can be either paired or unpaired with another entry in the log
* Adds `u32` time resolution and `u16` wall clock id fields to the report protocol


README.md
```diff
@@ -128,13 +130,39 @@ also include a binary payload denoting whether or not the expectation
...
     tx_status.is_ok(),
     "Measurement producer send result status",
     tags!("producer", "SEVERITY_10")
-)?;
+);
+```
+
+### Recording Wall Clock Time
+
+Wall clock time can be recorded as a standalone timestamp or
```
ZackPierce, line 144: Could we add a few words about recommended best practices for clock behavior? E.G. "Please make these numbers go up, and not down if you're faking it." E.G. "Use a consistent clock-source for all the time-related measurements at a given probe (don't mix and match clock-sources for a probe).

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -640,8 +743,21 @@ impl TryFrom<&[u8]> for Report {
                                 .interpret_as_event_id()
                                 .ok_or_else(|| SerializationError::InvalidEventId(raw_entry))?,
                         );
+                    } else if raw_entry.has_wall_clock_time_bits_set() {
```
ZackPierce, line 746: Don't we need to check this branch first?  Since the other two branches check a subset?

ZackPierce: Ah, I see this is not needed now that the `has_clock_bit_set` and `has_event_with_payload_bit_set` methods have been updated to really mean `has_*_signature_in_reserved_bits`.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -399,6 +402,12 @@ impl Collector {
     /// Collect all new logs, return a report
     pub fn collect_report(&mut self) -> Result<Option<Report>, Error> {
         self.rbuf.clear();
+
+        // Insert paired wall clock time entry from previous report if needed
+        if let Some(e) = self.prev_paired_wall_clock_time.take() {
+            self.rbuf.push(e);
```
ZackPierce, line 408: I don't think this works correctly in the face of missing events.  If we discover in the next line that some entries were missed, then that means we've lost the appropriate-to-pair entry (and currently now associating this cached time entry with some random entry later on).

ZackPierce: I think in order to get this right, you may have to write a test that creates conditions wherein the paired content gets overwritten, then we correctly discover this fact on the next read and decide to drop the cached wall-clock-time (WCT) entry rather than pairing it spuriously.

ZackPierce: An alternative approach when we detect that the paired content has been missed would be to downgrade the paired WCT into an unpaired WCT entry and keep it around.

collectors/modality-probe-collector-common/src/lib.rs
```diff
@@ -93,17 +99,70 @@ pub struct Report {
...
     pub frontier_clocks: Vec<LogicalClock>,
     pub event_log: Vec<EventLogEntry>,
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
 pub enum EventLogEntry {
-    Event(EventId),
-    EventWithPayload(EventId, u32),
-    TraceClock(LogicalClock),
+    Event(Option<Nanoseconds>, EventId),
```
ZackPierce, line 110: This is optional, and stylistic, but I would softly recommend flattening out the Option<Nanoseconds> by adding 3 new variants instead `TimedEvent`, `TimedEventWithPayload`, and `TimedTraceClock`.  It would reduce code motion around existing handling of specific variants and firmly point to more direct un-nested handling of each entry.

fenced-ring-buffer/src/lib.rs
```diff
@@ -179,6 +179,19 @@ where
...
+    pub fn first_entry(&self) -> &E {
+        match self {
+            Self::Single(e) => e,
+            Self::Double(e, _) => e,
+        }
+    }
+
+    /// Returns true if this is a double entry
+    pub fn is_double(&self) -> bool {
+        self.size() != 1
```
ZackPierce, line 193: A more direct and cheaper way to express this function body would be `matches!(self, WholeEntry::Double(_, _))`

modality-probe-cli/src/manifest_gen/c_parser.rs
```diff
@@ -138,6 +143,70 @@ fn parse_record_event_call_exp(input: Span) -> ParserResult<Span, EventMetadata>
...
+    let tags_pos = tags_and_desc.iter().position(|s| s.contains("tags="));
+    let mut tags = tags_pos
+        .map(|index| tags_and_desc.swap_remove(index))
+        .map(|s| s.replace("tags=", ""));
+    if let Some(t) = &mut tags {
+        if t.is_empty() {
+            return Err(make_failure(input, Error::EmptyTags(pos.into())));
+        }
+        if !t.contains("TIME") {
+            t.insert_str(0, "TIME;");
```
ZackPierce, line 191: So we're adding a `TIME` tag even if it isn't manually specified? I suppose this is alright if we document it somewhere.

src/error.rs
```diff
@@ -41,6 +41,58 @@ impl Display for InvalidProbeId {
...
+///
+/// Wall clock time must not be greater than Nanoseconds::MAX
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct InvalidWallClockTime;
+
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidWallClockTime {}
+
+#[cfg(feature = "std")]
+impl Display for InvalidWallClockTime {
```
ZackPierce, line 54: Looks like this impl of `Display` is compatible with the features available in `core::fmt`, probably could ease up on the std-flagging here with a std->core module rename.

collectors/modality-probe-debug-collector/src/lib.rs
jonlamb-gh: Good catch, will revisit this.

collectors/modality-probe-collector-common/src/lib.rs
jonlamb-gh: Yeah I think that makes sense too, will get this updated.

modality-probe-cli/src/manifest_gen/c_parser.rs
jonlamb-gh: Was on autopilot here, will remove and consider re-adding once we've had a chance to discuss any sort of time related auto-tagging capabilities.

collectors/modality-probe-debug-collector/src/lib.rs
```diff
@@ -399,20 +402,40 @@ impl Collector {
...
             let (ev, payload) =
                 LogEntry::event_with_payload(EventId::EVENT_LOG_ITEMS_MISSED, num_missed_rounded);
             self.rbuf.insert(0, WholeEntry::Double(ev, payload));
+
+            if let Some(e) = self.prev_paired_wall_clock_time.take() {
+                eprintln!("Warning: dropping paired wall clock time entry {:?} from previous report because items were missed and the associated entry was lost", e);
+            }
+        } else {
+            // Insert paired wall clock time entry from previous report if needed
+            if let Some(e) = self.prev_paired_wall_clock_time.take() {
```
ZackPierce, line 419: Can collapse this into the above `else` clause. I'm a little surprised `clippy` isn't demanding as much.

src/history.rs
```diff
@@ -224,6 +253,68 @@ impl<'a> DynamicHistory<'a> {
...
+    #[inline]
+    fn process_overwritten_log_entries(
+        &mut self,
+        first_overwritten: Option<WholeEntry<LogEntry>>,
+        second_overwritten: Option<WholeEntry<LogEntry>>,
+    ) {
+        self.merge_overwritten_clock(first_overwritten);
+        self.merge_overwritten_clock(second_overwritten);
+
+        // Fenced-ring-buffer loosed its missed count once we start pop'ing entries out
```
ZackPierce, line 269: typo? loosed -> lost / loses?

src/history.rs
```diff
@@ -531,13 +739,14 @@ impl<'a> DynamicHistory<'a> {
         Ok(())
     }
 
+    // NOTE: if there was an associated paired wall clock time entry
+    // (via snapshot merge/produce_with_time), it'll precede the logical clocks
```
ZackPierce, line 743: Seems like this information belongs on the `merge_internal` function as well -- the fact that in the log the order will end up something like "(wct)?local,remote"

src/lib.rs
```diff
@@ -471,11 +639,6 @@ impl PartialOrd for ModalityProbeInstant {
     }
 }
 
-/// Raw bytes related to extension metadata stored alongside
-/// the messages transmitted in this system (reports).
-#[derive(Debug)]
-pub struct ExtensionBytes<'a>(pub &'a [u8]);
```
ZackPierce, line 477: Good clean-up catch.

src/time.rs
```diff
@@ -0,0 +1,278 @@
...
+
+/// Nanosecond time resolution
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
+#[repr(transparent)]
+pub struct NanosecondResolution(pub u32);
+
+/// Nanosecond high bits
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
+#[repr(transparent)]
+pub struct NanosecondsHighBits(pub u32);
```
ZackPierce, line 22: If the NanosecondsXBits types are supposed to contain opaque bytes for transport (in a fixed endian-ness), I recommend we represent them as [u8;4] instead of u32.

ZackPierce: These should probably be accompanied by a few helper converters (constructors and extractors) that are explicit about their conversion input and output formats.

src/time.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    /// One seconds worth of nanoseconds
+    pub const SECOND: Self = Nanoseconds(1_000_000_000);
+    /// One milliseconds worth of nanoseconds
+    pub const MILLISECOND: Self = Nanoseconds(1_000_000);
+    /// One microseconds worth of nanoseconds
+    pub const MICROSECOND: Self = Nanoseconds(1_000);
+
+    /// Create a zero value time
+    pub const fn zero() -> Self {
```
ZackPierce, line 48: May consider making this a `const ZERO: Nanoseconds` as well for ease of comparison.

src/time.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl From<NanosecondsLowBits> for u32 {
+    fn from(bits: NanosecondsLowBits) -> Self {
+        bits.0
+    }
+}
+
+impl WallClockId {
+    /// This time domain is local to the probe only
+    pub const fn local_only() -> Self {
```
ZackPierce, line 196: This one deserves a `const LOCAL_ONLY: WallClockId`, too, I think.  Maybe an `is_local_only() -> bool` too if you're feeling generous.

ZackPierce (approved):
:+1:

PR #293, opened 2020-10-29
--radius switch for log CLI

### LINKED ISSUE #280: `--radius` filter for `log` cli
Provide a coordinate and a distance, and use that to filter the log output. Depends on #278.

pittma:
Closes #280.

### DIFF
--- a/modality-probe-cli/src/log/mod.rs
+++ b/modality-probe-cli/src/log/mod.rs
@@ -26,38 +26,6 @@ use radius::Radius;
 const COL_SPACE: &str = "   ";
 const COL_EDGE: &str = "---";
 
-const FORMAT_DOC: &str = "Provide a custom format string to be interpreted by each event row.
-
-The format string may use any combination of the following identifiers.
-
-| Specifier | Data               |
-|-----------|--------------------|
-|    %ei    | Event id           |
-|    %en    | Event name         |
-|    %ef    | Event file         |
-|    %el    | Event line         |
-|    %et    | Event tags         |
-|    %ed    | Event description  |
-|    %et    | Event type hint    |
-|    %ep    | Event payload      |
-|    %er    | Raw event payload  |
-|    %ec    | Event coordinate   |
-|    %eo    | Event clock offset |
-|    %pi    | Probe id           |
-|    %pn    | Probe name         |
-|    %pc    | Probe clock        |
-|    %pd    | Probe description  |
-|    %pf    | Probe file         |
-|    %pl    | Probe line         |
-|    %pt    | Probe tags         |
-|    %ci    | Component id       |
-|    %cn    | Component name     |
-|    %rt    | Receive time       |
-
-NOTE: If an identifier is used in the string and that field is not available on the event, it will be replaced by an empty string.
-
-";
-
 /// View the trace as a log.
 #[derive(Debug, PartialEq, StructOpt)]
 pub struct Log {
@@ -84,11 +52,54 @@ pub struct Log {
     #[structopt(short, parse(from_occurrences))]
     pub verbose: u8,
 
-    #[structopt(short, long, help = FORMAT_DOC)]
+    /// Provide a custom format string to be interpreted by each event
+    /// row.
+    ///
+    /// The format string may use any combination of the following
+    /// identifiers.
+    ///
+    /// | Specifier | Data               |
+    /// |-----------|--------------------|
+    /// |    %ei    | Event id           |
+    /// |    %en    | Event name         |
+    /// |    %ef    | Event file         |
+    /// |    %el    | Event line         |
+    /// |    %et    | Event tags         |
+    /// |    %ed    | Event description  |
+    /// |    %et    | Event type hint    |
+    /// |    %ep    | Event payload      |
+    /// |    %er    | Raw event payload  |
+    /// |    %ec    | Event coordinate   |
+    /// |    %eo    | Event clock offset |
+    /// |    %pi    | Probe id           |
+    /// |    %pn    | Probe name         |
+    /// |    %pc    | Probe clock        |
+    /// |    %pd    | Probe description  |
+    /// |    %pf    | Probe file         |
+    /// |    %pl    | Probe line         |
+    /// |    %pt    | Probe tags         |
+    /// |    %ci    | Component id       |
+    /// |    %cn    | Component name     |
+    /// |    %rt    | Receive time       |
+    ///
+    /// NOTE: If an identifier is used in the string and that field is not
+    /// available on the event, it will be replaced by an empty string.
+    #[structopt(short, long, verbatim_doc_comment)]
     pub format: Option<String>,
+
     /// Filter a whole graph down to the radius around a specific
     /// event.
-    #[structopt(long)]
+    ///
+    /// Radius takes an event coordinate and the length of the radius
+    /// around that coordinate to be included in the output.
+    ///
+    /// It takes the form:
+    ///
+    /// <coord>,<radius>
+    ///
+    /// Where <coord> is a colon separated event coordinate such as
+    /// 1:2:3:4.
+    #[structopt(long, verbatim_doc_comment)]
     pub radius: Option<String>,
 }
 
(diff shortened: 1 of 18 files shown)


modality-probe-cli/src/log/mod.rs
```diff
@@ -82,6 +86,10 @@ pub struct Log {
 
     #[structopt(short, long, help = FORMAT_DOC)]
     pub format: Option<String>,
+    /// Filter a whole graph down to the radius around a specific
+    /// event.
```
jonlamb-gh, line 90: I think it'd be helpful here or in the about/help to mention what the units of `--radius` mean.

jonlamb-gh: ah, I see, it's now about a specific event.

modality-probe-cli/src/log/mod.rs
```diff
@@ -222,6 +234,7 @@ fn print_as_graph<W: WriteIo>(
     mut stream: W,
 ) -> Result<(), Box<dyn std::error::Error>> {
     // Edges waiting to be drawn.
+    // TODO: use increment / prev for these before moving on to testing.
```
jonlamb-gh, line 237: Is this something we need a ticket for?

modality-probe-cli/src/log/mod.rs
jonlamb-gh: Can you add some guidance or examples somewhere in the README or here in this help about how to use `--radius`?

nathanaschbacher:
I don't know if Dan captured this anywhere, but we spoke, and radius is
only supposed to be a single number not a compound thing with a coordinate
and a radius. The coordinate should be specified separately in a --coord or
--coordinate or --at or something arg.

jonlamb-gh:
> I don't know if Dan captured this anywhere, but we spoke, and radius is only supposed to be a single number not a compound thing with a coordinate and a radius. The coordinate should be specified separately in a --coord or --coordinate or --at or something arg.

Ah I see. Yeah I agree, `--radius` to me feels like it should be a single dimension akin to distance, starting from some coordinate.

modality-probe-cli/src/log/radius.rs
```diff
@@ -0,0 +1,407 @@
...
+        }
+        let coord = sections[0];
+        let c_sections: Vec<&str> = coord.split(':').collect();
+        if c_sections.len() != 4 {
+            give_up!(format!("{} is not a valid coordinate", coord));
+        }
+
+        let distance = hopefully!(
+            sections[1].parse::<usize>(),
+            "Unable to parse the given distance"
```
jonlamb-gh, line 39: Can we make the coordinate sub-component parser errors unique, so that we can tell the user which part of the supplied coordinate was problematic?

src/lib.rs
```diff
@@ -223,6 +221,28 @@ impl LogicalClock {
...
 
+    /// Calculate the clock which follows this one.
+    pub fn next(&self) -> LogicalClock {
+        let mut new = *self;
+        new.increment();
+        new
+    }
+
+    /// Calculate the clock which preceeded this one.
+    pub fn prev(&self) -> LogicalClock {
```
jonlamb-gh, line 232: Could you add some testing around the next/prev mechanisms here showing they hold up across the overflow boundaries?

modality-probe-cli/src/log/mod.rs
```diff
@@ -335,14 +348,13 @@ fn print_as_graph<W: WriteIo>(
...
                         // This is a local clock.
-                        let (lc_id, clock) = lc.pack();
                         if lc.id == *probe_id {
                             // The self-clock `0` is useful from a
                             // causal standpoint, but from a
                             // coordinating-timelines standpoint, it's
                             // ambiguous because it appears whether a
                             // snapshot was produced or not.
-                            if clock != 0 {
+                            if lc.pack().1 != 0 {
```
jonlamb-gh, line 357: Will this logic be problematic for probes that implement restart tracking?

nathanaschbacher:
I think there might generally be some work that needs to be done to deal
with everything related to weak relationships. Like, restart tracking,
missing reports, cross run/session boundaries, etc.

modality-probe-cli/src/log/mod.rs
pittma: Yeah, I intend to make a docs pass this morning.

modality-probe-cli/src/log/mod.rs
pittma: It's actually done! I just forgot to remove the TODO.

pittma:
> > I don't know if Dan captured this anywhere, but we spoke, and radius is only supposed to be a single number not a compound thing with a coordinate and a radius. The coordinate should be specified separately in a --coord or --coordinate or --at or something arg.
> 
> Ah I see. Yeah I agree, `--radius` to me feels like it should be a single dimension akin to distance, starting from some coordinate.

I haven't circled back to this yet, no. I find the idea of requiring two switches awkward because it's not clear to me what either would do if used on their own. If `--radius` on its own is confusing / surprising, I need to think a bit more about how better to present it.

modality-probe-cli/src/log/mod.rs
pittma: IIUC, no, I don' t _think_ so. This clock is packed, so a `0` would be

```rust
LogicalClock {
    id,
    epoch: ProbeEpoch(0),
    ticks: ProbeTicks(0),
}
```

jonlamb-gh:
> > > I don't know if Dan captured this anywhere, but we spoke, and radius is only supposed to be a single number not a compound thing with a coordinate and a radius. The coordinate should be specified separately in a --coord or --coordinate or --at or something arg.
> > 
> > 
> > Ah I see. Yeah I agree, `--radius` to me feels like it should be a single dimension akin to distance, starting from some coordinate.
> 
> I haven't circled back to this yet, no. I find the idea of requiring two switches awkward because it's not clear to me what either would do if used on their own. If `--radius` on its own is confusing / surprising, I need to think a bit more about how better to present it.

I wonder if we could lean on `StructOpt` for this using `required_if` or `required_unless` attributes to help ensure they're supplied together?

pittma:
> I think there might generally be some work that needs to be done to deal with everything related to weak relationships. Like, restart tracking, missing reports, cross run/session boundaries, etc.

* Yes on multiple sessions in a single trace file.
* Restart tracking, I don't think so, unless there's something probe side that needs to be done. The `visualize` and `log` code just follows the clocks which are meant to contain restart information, right?
* I'm not sure what you mean about missing reports, are you talking about enumerating holes in the graph? `log` does that, but `visualize` does not.

pittma:
> I wonder if we could lean on StructOpt for this using required_if or required_unless attributes to help ensure they're supplied together?..

Oh totally, I know we _can_ enforce that they be used together, but what would the point be of their being two different arguments if they both had to be used together?

jonlamb-gh:
> > I wonder if we could lean on StructOpt for this using required_if or required_unless attributes to help ensure they're supplied together?..
> 
> Oh totally, I know we _can_ enforce that they be used together, but what would the point be of their being two different arguments if they both had to be used together?

That's a good point. I'm not sure which is better from the user's perspective here, I can see it working both ways. :shrug: 
I think I'm personally biased towards separated because I like the idea of radius being a distance of sorts, and coordinates being a location. Maybe it makes it easier to copy/paste coordinates around while leaving radius up to the invocation at hand, but then again I'm not sure how strong a use case that is - I think what you've laid out could accomplish the same.

modality-probe-cli/src/log/radius.rs
jonlamb-gh: Nice, I like it.

nathanaschbacher:
They're probably both required to supply all the time. The command is
invalid with both parts, but I don't think that means we need to create a
special secondary syntax that glues the parts together under one command.
Lots of CLI tools have dependent arguments where you have to supply all the
dependent parts for any of them to work, and it's an error or suitable
default behavior + warning when they're not supplied (we may not have a
suitable default behavior here).

I could see there being a --from flag that takes a coordinate. So you could
have a --radius X --from C, and separately have a --future --from C,
--history --from C. To a certain extent --radius also then acts like a
limiter modifier on --history & --future which basically tells --radius to
only walk one direction. Really you need a way to say "some spot", that's
the most important thing, and then everything else is a command that
modifies the behavior of what to do relative to that spot.

nathanaschbacher:
*is invalid without both parts

jonlamb-gh (approved):

PR #308, opened 2020-12-03
Move test step after fmt/clippy

jonlamb-gh:


pittma (approved):
What a wonderful idea!

PR #317, opened 2021-01-15
ending a report on a clock is okay if it isn't followed by a foreign clock

### LINKED ISSUE #315: A report can't end with a local clock, even if that clock is a result of snapshot production
I _think_ this https://github.com/auxoncorp/modality-probe/blob/c9e1ebf92aa5ff5a6ae5fc9117ca8ad25e2d0a83/src/history.rs#L577-L581

Is making a probe which produces a snapshots but isn't recording events unable to get a report out. It should probably peek ahead to see if the next entry is both a double entry and a foreign clock.

pittma:
Closes #315.

### DIFF
--- a/tests/integration_tests.rs
+++ b/tests/integration_tests.rs
@@ -281,6 +281,81 @@ fn report_buffer_too_small_error() -> Result<(), ModalityProbeError> {
         raw_event,
         EventId::EVENT_INSUFFICIENT_REPORT_BUFFER_SIZE.get_raw()
     );
+    Ok(())
+}
+
+#[test]
+fn report_can_end_in_local_clock_from_snapshot_produce() -> Result<(), ModalityProbeError> {
+    let mut storage = [MaybeUninit::new(0u8); 512];
+    let probe = ModalityProbe::try_initialize_at(
+        &mut storage,
+        1,
+        NanosecondResolution::UNSPECIFIED,
+        WallClockId::local_only(),
+        RestartCounterProvider::NoRestartTracking,
+    )?;
+    let _snap = probe.produce_snapshot();
+    // Room for the zero-clock, the initialized event, and the
+    // produced snapshot.
+    let mut report_dest = vec![0u8; wire::WireReport::<&[u8]>::buffer_len(1, 5)];
+    let bytes_written = probe.report(&mut report_dest)?.unwrap();
+    // We should have written the frontier clock, the initialized
+    // event, and the pair for the produced snapshot.
+    assert_eq!(
+        bytes_written.get(),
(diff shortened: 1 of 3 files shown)


jonlamb-gh:
Good catch on the report foreign clock fragmentation bug.

IIUC, these changes affect how entries are ordered when
they're put back into the log.

In the case where we have to bail out early due to limited user buffer size,
a log entry `pop`'d off the tail (oldest in our fenced-ring-buffer impl) is pushed back onto the head (newest).

For example, if I were to log a few events like so:

```rust
probe.record_event(EventId::new(1).unwrap());
probe.record_event(EventId::new(2).unwrap());
probe.record_event(EventId::new(3).unwrap());
```

Using a limited report buffer size, like `wire::WireReport::<&[u8]>::buffer_len(1, 4)`,
will require 2 reports to drain.

With this change, I'll get something like this in the report buffers:

```
# 1st report
event_log: [
    TraceClock( ... ),
    Event(EventId(1073741817)),
    Event(EventId(1)),
]

# 2nd report
event_log: [
    Event(EventId(3)),
    Event(EventId(2)),
    Event(EventId(1073741822)),
]
```

The events are now reordered `1, 3, 2`, whereas with the previous `peek` mechanics
the ordering is preserved `1, 2, 3`.

Was this the desired behavior?

pittma:
Shoot. Nope. I did not know that that was how pop / push behaved on the fenced ring. That's bad. I'll figure out a way to work around that tomorrow. No tests to catch this is weird too.

tests/integration_tests.rs
```diff
@@ -501,12 +577,21 @@ proptest! {
             log_bytes[tail_index - 1],
         ]);
         let second_to_last_item = unsafe { log::LogEntry::new_unchecked(second_to_last_item) };
+	println!("{}", log_report.n_log_entries());
```
jonlamb-gh, line 580: Is this println vestigial?

tests/integration_tests.rs
```diff
@@ -501,12 +577,21 @@ proptest! {
...
                 second_to_last_item.interpret_as_logical_clock_probe_id(),
-                snap_probe_id
+		// In the case where a report buffer is big enough to even one
+		// interaction set, the first expression of this conjunction
+		// is still accidentally true—it's the initial self clock @
+		// 0. In that case it should == probe_id not snap_probe_id.
+		if log_report.n_log_entries() == 3 {
+		    probe_id
+		} else {
+                    snap_probe_id
```
jonlamb-gh, line 593: Indentation got a little wonky in this test

jonlamb-gh (approved):

PR #320, opened 2021-02-13
Install system packages in weekly_fuzz workflow needed by debug collector

jonlamb-gh:
Fixes https://github.com/auxoncorp/modality-probe/runs/1892693303?check_suite_focus=true


jonlamb-gh:
Will follow up with new clippy changes and a version bump.

PR #343, opened 2021-08-10
Fenced ring buffer more strongly associates wall clock times with buddy entries

ZackPierce:
More specifically, we more deeply support the tight relationship between wall clock time entry-pairs and their following content.
* Adds `Triple` and `Quad` variants of `WholeEntry`. These cover the paired WCT + something-else variations
* Now that the FRB's `WholeEntry` can more completely cover the WCT cases, reduce some of the double-WholeEntry peeking and special casing.
* The effects necessary to continue the extremely thin layer of abstraction that the low-level `Entry` trait is providing over the `LogEntry` data format is papered with tongue-in-cheek naming for the relevant concepts, e.g. `is_mega_variable_prefix` means "the kind of `Entry` that signals that it precedes both a certainly-following entry with more content but also one OR two further entries, which we won't be able to know how many of until we read more.
* Rename `has_clock_bit_set` to `has_logical_clock_bit_set` for clarity.

### DIFF
--- a/fenced-ring-buffer/src/async_reader.rs
+++ b/fenced-ring-buffer/src/async_reader.rs
@@ -49,6 +49,10 @@ where
     storage_cap: usize,
     /// Vec used to store temporary snapshot of new entries in the buffer
     buf_snapshot: Vec<E>,
+    /// Cached mega-prefix, not put into buffer until suffix is successfully read
+    stored_mega_prefix: Option<E>,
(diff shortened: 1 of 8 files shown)


jonlamb-gh (commented):
Changes look good.
Looks like we still need to figure out what to do with the
fenced-ring-buffer TLA+ model bits and its README.

PR #348, opened 2022-06-28
Pin cross to 0.2.1

jonlamb-gh:


jonlamb-gh:
Going to investigate the failing cross tests in https://github.com/auxoncorp/modality-probe/runs/7090286481?check_suite_focus=true before pinning versions

